PR #25, opened 2020-06-02
new orient seqs method

### LINKED ISSUE #9: add method for orienting mixed-orientation reads
Some ideas and features (from conversation with @mikerobeson ):

* need util function to detect orientation — this could be used as a stand-alone validator?
* Add reverse-complement option
* Need to handle IUPAC codes appropriately in reverse complementation (skbio may have a built-in to do this?)

The workflow would be:
1. iterate over seqs in fasta
2. detect orientation (relative to reference sequences? kmer-based approach for quick estimation?)
3. fix orientation (as needed)
4. write out

nbokulich:
fixes #9 

@mikerobeson doesn't quite hit everything we discussed in #9 (e.g., making a stand-alone utility to check orientation) but it works (if the tests are to be believed... if you're up for it it could use some live testing to make sure it hits the corners... have any datasets in which you know the orientations are mixed???)


thermokarst:
This PR should've included vsearch as a conda dependency, I have addressed that here:

https://github.com/bokulich-lab/RESCRIPt/pull/139

PR #52, opened 2020-06-24
added method to download, parse, and import data from NCBI

BenKaehler:
Added method to download, parse, and import data from NCBI.


rescript/ncbi.py
```diff
@@ -0,0 +1,141 @@
...
+    return seqs, taxids
+
+
+def get_taxonomies(taxids):
+    params = dict(db='taxonomy')
+    ids = list(map(str, taxids.values()))
+    records = get_from_post(ids, params)
+    taxa = {}
+    levels = ['kingdom', 'phylum', 'class', 'order', 'family', 'genus',
+              'species']
```
mikerobeson, line 119: We should likely grab all taxonomies and then subset / propagate them afterwards. We also applied this approach to the [SILVA parsing](https://github.com/bokulich-lab/RESCRIPt/blob/772294cd0a9fed069f2b2dfcd0f1d6144722eff3/rescript/parse_silva_taxonomy.py#L23).  

I suggest this, as many eukaryote data in GenBank do not have , e.g. an "Order" or "Family" but have "Sub-Order" or "Super-Family"... or some other rank. This was always an issue for me when building databases for other marker genes, like 12S, COI, trnL. Especially when plants have "Divisions", "Tribes", and "Varieties". 

Anyway... this is all assuming we want to open this up beyond microbial taxonomy. I figured, it should be on our radar, even if only for a later release.  :-)

rescript/plugin_setup.py
```diff
@@ -634,6 +635,31 @@
...
+    function=get_ncbi_data,
+    inputs={},
+    parameters={
+        'query': Str,
+        'accession_ids': Metadata},
+    outputs=[('sequences', FeatureData[Sequence]),
+             ('taxonomy', FeatureData[Taxonomy])],
+    input_descriptions={},
+    parameter_descriptions={
+        'query': 'Query on the NCBI Nucleotide database',
```
mikerobeson, line 648: Perhaps a short example here? Like: "txid1239[ORGN] NOT environmental[Title]", or whatever your favorite search is. Or perhaps add a links to these:

- https://www.ncbi.nlm.nih.gov/books/NBK49540/
- https://www.ncbi.nlm.nih.gov/books/NBK3837/

There maybe better ones...

mikerobeson:
I tried to be brutal and use this overly complex query to download Actinopterygii COI data.  :-P

```
qiime rescript get-ncbi-data --p-query "txid7898[ORGN] AND (cytochrome c oxidase subunit 1[Title] OR cytochrome c oxidase subunit I[Title] OR cytochrome oxidase subunit 1[Title] OR cytochrome oxidase subunit I[Title] OR COX1[Title] OR CO1[Title] OR COI[Title]) AND mitochondrion[Filter] NOT environmental sample[Title] NOT environmental samples[Title] NOT environmental[Title] NOT uncultured[Title] NOT unclassified[Title] NOT unidentified[Title] NOT unverified[Title]" --o-sequences ~/Desktop/co1-actinopterygii-seqs.qza --o-taxonomy ~/Desktop/co1-actinopterygii-tax.qza --verbose
```

I received this error:
```
Traceback (most recent call last):
  File "/Users/robesonmichael/AnalysisSoftware/miniconda3/envs/qiime2-dev/lib/python3.6/site-packages/q2cli/commands.py", line 328, in __call__
    results = action(**arguments)
  File "<decorator-gen-529>", line 2, in get_ncbi_data
  File "/Users/robesonmichael/AnalysisSoftware/miniconda3/envs/qiime2-dev/lib/python3.6/site-packages/qiime2/sdk/action.py", line 245, in bound_callable
    output_types, provenance)
  File "/Users/robesonmichael/AnalysisSoftware/miniconda3/envs/qiime2-dev/lib/python3.6/site-packages/qiime2/sdk/action.py", line 390, in _callable_executor_
    output_views = self._callable(**view_args)
  File "/Users/robesonmichael/AnalysisSoftware/github.forks/RESCRIPt/rescript/ncbi.py", line 36, in get_ncbi_data
    taxa = get_taxonomies(taxids)
  File "/Users/robesonmichael/AnalysisSoftware/github.forks/RESCRIPt/rescript/ncbi.py", line 137, in get_taxonomies
    taxonomy = taxa[taxid]
KeyError: '352920'

Plugin error from rescript:

  '352920'
```

I ran into this quite often when I used to regularly download and parse GenBank nucleotide data. In the past I've had to manually update the taxids... or simply bypass this with a warning for records with this taxid. Often, *though apparently not in this case*, the taxid was updated but not reflected in the taxonomy files that you can download from genbank. I used to determine this by manually looking up the taxid on the GenBank Taxonomy page (again this seems fine in this case). 

Probably a good idea to pass these failed instances with an error and/or write out to a file with all the records that failed to be parsed? Unless there is something else I am missing?

Anyway, it is quite often for the taxonomy information and sequences to be out of sync when it comes to taxids.

rescript/ncbi.py
```diff
@@ -0,0 +1,141 @@
...
+        DNAIterator, DataFrame):
+    if query is None and accession_ids is None:
+        raise ValueError('Query or accession_ids must be supplied')
+
+    if query:
+        seqs, taxids = get_nuc_for_query(query)
+
+    if accession_ids:
+        accs = accession_ids.get_ids()
+        if query and seqs:
```
nbokulich, line 27: should we just allow a query or list of accession IDs instead of a mix of both?

rescript/ncbi.py
```diff
@@ -0,0 +1,141 @@
...
+            if accs:
+                acc_seqs, acc_taxids = get_nuc_for_accs(accs)
+                seqs.update(acc_seqs)
+                taxids.update(acc_taxids)
+        else:
+            seqs, taxids = get_nuc_for_accs(accs)
+
+    taxa = get_taxonomies(taxids)
+
+    seqs = DNAIterator(DNA(v, metadata={'id': k}) for k, v in seqs.items())
```
nbokulich, line 38: why create a DNAIterator? You could create an empty DNAFASTAFormat object, then just parse and save seqs to that as they are generated

rescript/tests/test_ncbi.py
```diff
@@ -0,0 +1,36 @@
...
+        df = DataFrame(index=['M59083.2', 'AJ234039.1'])
+        df.index.name = 'id'
+        md = Metadata(df)
+        get_ncbi_data = rescript.methods.get_ncbi_data
+        acc_seq, acc_tax = get_ncbi_data(accession_ids=md)
+        que_seq, que_tax = get_ncbi_data(query='M59083.2 OR AJ234039.1')
+
+        acc_seq = {s.metadata['id']: str(s) for s in acc_seq.view(DNAIterator)}
+        que_seq = {s.metadata['id']: str(s) for s in que_seq.view(DNAIterator)}
+        self.assertEqual(acc_seq, que_seq)
```
nbokulich, line 29: could you add some expected results and then confirm that both acc and query methods retrieve those expected results? e.g., what if both queries are failing and output empty files... would these tests pass?

rescript/tests/test_ncbi.py
```diff
@@ -0,0 +1,36 @@
...
+
+        acc_tax = acc_tax.view(DataFrame).to_dict()
+        que_tax = que_tax.view(DataFrame).to_dict()
+        self.assertEqual(acc_tax, que_tax)
```
nbokulich, line 33: if viewing as a dataframe, could you use the pandas testing functions to test? It's more informative than `assertEqual` if there is a mismatch.

rescript/tests/test_ncbi.py
```diff
@@ -0,0 +1,36 @@
...
+
+        self.assertEqual(len(acc_seq), 2)
```
nbokulich, line 35: you could drop these final 2 lines if you compare against some expected results... that would be more robust than counting the number of records. Thanks!

rescript/plugin_setup.py
nbokulich: I like both of these ideas... and you could add this to the `description` instead of here if you decide to add an extended description.

nbokulich: alternatively, maybe we should create a little tutorial/cookbook in the README.md file and just direct users to the repo for more details (a cookbook would be really useful for common queries like the 16S refseqs, and would help me test this PR too)

rescript/ncbi.py
BenKaehler: We've had this conversation offline, but for posterity, there is a use case here.

What if you wanted to download the RefSeq Targeted Loci for 16S (query: '33175[BioProject] OR 33317[BioProject]') supplemented with all the sequences they used for the Stirrups custom DB (https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3535711/bin/1471-2164-13-S8-S17-S4.xls)?

rescript/ncbi.py
BenKaehler: I agree that this needs to be configurable in some way. I am also happy to leave it to later.

As a first step, I have made "levels" an input parameter.

We could also add an "allowed-levels" an input parameter so that we could provide your silva level extrapolation algorithm.

To help me understand your algorithm, do you build a taxonomic tree based on "allowed-levels", of which the "levels" are a subset, then propagate labels down branches until all levels in "levels" are filled?

rescript/plugin_setup.py
BenKaehler: Why don't we add it to the existing [tutorial](https://forum.qiime2.org/t/processing-filtering-and-evaluating-reference-sequence-data-with-rescript/15494)?

rescript/ncbi.py
BenKaehler: I did it that way because that's the way I've always returned something to be saved as a FeatureData[Sequence]. Is creating a DNAFASTAFormat object preferable?

rescript/tests/test_ncbi.py
BenKaehler: I tried that first but the queries don't necessarily return the taxonomies in the same order.

rescript/tests/test_ncbi.py
BenKaehler: as you wish

rescript/tests/test_ncbi.py
BenKaehler: as you wish

BenKaehler:
Why can't I reply to [this comment](https://github.com/bokulich-lab/RESCRIPt/pull/52#issuecomment-649066055)?

My reply is that I have added a warning for when this happens and made it so it doesn't break.

rescript/ncbi.py
mikerobeson: Hi Ben,

The literal taxonomy tree was provided by SILVA. I just traversed the tree and pulled the taxids from all the nodes. Then used them to lookup the taxonomy and rank information from the other files they provided. However, this is not unlike what I've done with the old pycogent NCBI taxonomy parser code. I just traversed the hierarchy and either pulled everything.. or only the ranks I wanted and continued on. I can e-mail you some *rather crappy* bits of code (python 2.7) that I wrote to fetch genbank taxonomy from local blast output, based on 'alllowed ranks'.

rescript/tests/test_ncbi.py
mikerobeson: Couldn't you simply [sort by the index](https://github.com/mikerobeson/RESCRIPt/blob/e9b0c8d73cc46780765d191409f45ded602e518e/rescript/tests/test_parse_silva_taxonomy.py#L105) for both, and then compare them? or am I missing something?

rescript/plugin_setup.py
nbokulich: sure, adding existing examples to the existing tutorial works for now. Would be nice to get a dedicated tutorial/split up that monstrously long tutorial into thematic subjects... but that is not a prerequisite to merging this PR!

rescript/ncbi.py
nbokulich: Yes, I think `DNAFASTAFormat` is better here.

rescript/ncbi.py
mikerobeson: > you build a taxonomic tree based on "allowed-levels", of which the "levels" are a subset, then propagate labels down branches until all levels in "levels" are filled

Sorry, I forgot to address that. Essentially yes. Though in this case I was copying and propagating within a dataframe. But your cae case. I'd recommend propagation BEFORE extracting the levels you want. This way we would not have to worry about empty 'Family' or other empty ranks that might turn up. That is, the taxonomy would be propagated before you'd subset that information. This is what we are doing with SILVA via the dataframe.

mikerobeson:
So I re-ran the same query and received this different error:

>Traceback (most recent call last):
  File "/Users/robesonmichael/AnalysisSoftware/miniconda3/envs/qiime2-dev/lib/python3.6/site-packages/q2cli/commands.py", line 328, in __call__
    results = action(**arguments)
  File "<decorator-gen-529>", line 2, in get_ncbi_data
  File "/Users/robesonmichael/AnalysisSoftware/miniconda3/envs/qiime2-dev/lib/python3.6/site-packages/qiime2/sdk/action.py", line 245, in bound_callable
    output_types, provenance)
  File "/Users/robesonmichael/AnalysisSoftware/miniconda3/envs/qiime2-dev/lib/python3.6/site-packages/qiime2/sdk/action.py", line 390, in _callable_executor_
    output_views = self._callable(**view_args)
  File "/Users/robesonmichael/AnalysisSoftware/github.forks/RESCRIPt/rescript/ncbi.py", line 44, in get_ncbi_data
    taxa = get_taxonomies(taxids, levels, entrez_delay)
  File "/Users/robesonmichael/AnalysisSoftware/github.forks/RESCRIPt/rescript/ncbi.py", line 129, in get_taxonomies
    for level in rec['LineageEx']['Taxon']:
TypeError: string indices must be integers
Plugin error from rescript:
string indices must be integers
See above for debug info.

BenKaehler:
For future reference, Mike's post on label propagation: https://forum.qiime2.org/t/silva-138-classifiers/13131/26

BenKaehler:
I have added rank propagation.

Thanks @nbokulich and @mikerobeson for your thorough review and wise comments. I think I have now addressed most of your comments. Please let me know if I have missed anything.

(I may also make some changes when Evan gets back to me on proper usage of List(Str).)

rescript/ncbi.py
nbokulich: actually... what is `seqs` here? looks like it is a dict? in which case creating an iterator to save makes sense.

rescript/plugin_setup.py
```diff
@@ -641,19 +641,25 @@
     parameters={
         'query': Str,
         'accession_ids': Metadata,
-        'entrez_delay': Float,
-        'levels': List[Str]},
+        'ranks': List[Str],
+        'allowed_ranks': List[Str],
+        'disable_rank_propagation': Bool,
```
mikerobeson, line 646: Wouldn't it be easier to change this to `enable_rank_propagation` and `no_enable_propagation`? 
If feels odd to say the following as a double-negative : `no_disable_rank_propagation`. Also potentially harder for a user to understand / interpret which of those two choices is `False`.

rescript/ncbi.py
```diff
@@ -0,0 +1,191 @@
...
+def get_ncbi_data(
+        query: str = None, accession_ids: Metadata = None,
+        ranks: list = None, allowed_ranks: list = None,
+        disable_rank_propagation: bool = False,
+        entrez_delay: float = 0.334) -> (DNAIterator, DataFrame):
+    if query is None and accession_ids is None:
+        raise ValueError('Query or accession_ids must be supplied')
+    if ranks is None:
+        ranks = _default_ranks
+    if allowed_ranks is None:
```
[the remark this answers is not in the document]
mikerobeson: I second @nbokulich's idea.

rescript/plugin_setup.py
nbokulich: good call!

rescript/ncbi.py
BenKaehler: Thanks Mike for all your feedback (and chats offline). I took this algorithm and implemented in my taxonomy parsing code. I hope I got it right!

rescript/plugin_setup.py
BenKaehler: Good call. RESCRIPt is going to do a bunch of different things for different sets of users.

rescript/ncbi.py
BenKaehler: Thanks!

rescript/tests/test_ncbi.py
BenKaehler: Do you want me to change it to that?

rescript/plugin_setup.py
BenKaehler: I phrased it like that because `no_disable_rank_propagation` is the default and `no_enable_propagation` don't read good. So a user would only ever enter `--p-disable-rank-propagation`, which seems more self-explanatory.

But I can change it if you want.

rescript/ncbi.py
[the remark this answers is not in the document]
BenKaehler: Ok, but to do that we have to come up with an exhaustive list of the trusted `allowed_ranks` across all the clades of the tree of life, and we would have to update it whenever someone thinks of a new one.

For instance, the current list does not include "Divisions", "Tribes", and "Varieties".

Is it possible to get an exhaustive list?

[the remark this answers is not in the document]
nbokulich: if users really need some newfangled rank name, we can fix (or they can put in a PR)

I don't think there will be an infinite variety of rank names... the most common rank names will do for us I think.

I'd like to avoid the alternative, in which users run into all sorts of issues by specifying their own...

rescript/plugin_setup.py
nbokulich: how about `rank_propagation` and `no_rank_propagation`?

rescript/tests/test_ncbi.py
mikerobeson: I am fine either way.

rescript/plugin_setup.py
mikerobeson: Elegant and to-the-point. _We likes it precious. We likes it!_

BenKaehler: That's the winner. Thanks Nick.

rescript/ncbi.py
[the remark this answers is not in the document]
BenKaehler: That's a fair point. Do either of you have a slightly more exhaustive list than the current allowed_ranks? This conversation started because we wanted to make the system cater for more than microbes.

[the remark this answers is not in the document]
BenKaehler: Please select which of these you would like in the canonical allowed ranks:

Domain
Subdomain
Hyperkingdom
Superkingdom
Kingdom
Subkingdom
Infrakingdom
Parvkingdom
Superphylum
Phylum
Subphylum
Infraphylum
Microphylum
Superclass
Class
Subclass
Infraclass
Parvclass
Superdivision
Division
Subdivision
Infradivision
Superlegion
Legion
Sublegion
Infralegion
Supercohort
Cohort
Subcohort
Infracohort
Superorder
Gigaorder
Magnorder
Grandorder
Mirorder
SeriesFish
Order
Nanorder
Hypoorder
Suborder
Infraorder
Parvorder
Section
Subsection
Gigafamily
Megafamily
Grandfamily
Hyperfamily
Superfamily
Epifamily
SeriesLepidoptera
GroupLepidoptera
Family
Subfamily
Infrafamily
Supertribe
Tribe
Subtribe
Infratribe
Genus
Subgenus
SeriesBotany
SubseriesBotany
SpeciesGroup
SpeciesSubgroup
Species
Subspecies
Varietas
Subvarietas
Forma
Subforma
Cultivar
Breed
Individual
Unspecified

With thanks to https://docs.rs/taxonomy/0.3.3/taxonomy/enum.TaxRank.html

BenKaehler:
Note to self: include xmltodict in readme.

rescript/tests/test_ncbi.py
```diff
@@ -0,0 +1,55 @@
...
+        acc_seq, acc_tax = self.get_ncbi_data(accession_ids=md)
+        acc_seq = {s.metadata['id']: str(s) for s in acc_seq.view(DNAIterator)}
+        seqs = {s.metadata['id']: str(s) for s in self.seqs.view(DNAIterator)}
+        self.assertEqual(acc_seq, seqs)
+
+        acc_tax = acc_tax.view(DataFrame).to_dict()
+        taxa = self.taxa.view(DataFrame).to_dict()
+        self.assertEqual(acc_tax, taxa)
+
+    def test_get_ncbi_data_query(self):
```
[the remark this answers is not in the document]
BenKaehler: done

rescript/ncbi.py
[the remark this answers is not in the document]
BenKaehler: ok, I've hard coded the allowed ranks and made appropriate changes to the rank propagation code

rescript/ncbi.py
```diff
@@ -135,41 +146,55 @@ def get_nuc_for_query(query, entrez_delay=0.):
...
+            # if we care about genus and genus is in the species label and
+            # we don't already know genus, split it out
+            if 'genus' in ranks:
+                if taxonomy['genus']:
+                    if species.startswith(taxonomy['genus'] + ' '):
+                        species = species[len(taxonomy['genus']) + 1:]
+                elif ' ' in species:
+                    genus, species = species.split(' ', 1)
+                    taxonomy['genus'] = genus
+            taxonomy['species'] = species
```
mikerobeson, line 196: Nice!

rescript/tests/test_ncbi.py
```diff
@@ -43,6 +68,23 @@ def test_get_ncbi_data_accession_ids(self):
...
         self.assertEqual(acc_tax, taxa)
 
+    def test_get_ncbi_data_rank_propagation_nonstandard_ranks(self):
+        df = DataFrame(index=['M59083.2', 'AJ234039.1'])
+        df.index.name = 'id'
+        md = Metadata(df)
+
+        acc_seq, acc_tax = self.get_ncbi_data(
+                accession_ids=md,
+                ranks=['subkingdom', 'subphylum', 'subclass', 'suborder',
```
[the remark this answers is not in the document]
mikerobeson: I'm glad you caught that, as I missed it! You can follow the approach I took [here](https://github.com/bokulich-lab/RESCRIPt/blob/390a6162b6c43c81eba5135f8063180d067e9f91/rescript/parse_silva_taxonomy.py#L25).

>ALLOWED_RANKS = OrderedDict({'domain': 'd__', 'superkingdom': 'sk__',
                             'kingdom': 'k__', 'subkingdom': 'ks__',
                             'superphylum': 'sp__', 'phylum': 'p__',
                             'subphylum': 'ps__', 'infraphylum': 'pi__',
                             'superclass': 'sc__', 'class': 'c__',
                             'subclass': 'cs__', 'infraclass': 'ci__',
                             'superorder': 'so__', 'order': 'o__',
                             'suborder': 'os__', 'superfamily': 'sf__',
                             'family': 'f__', 'subfamily': 'fs__',
                             'genus': 'g__'})

More specifically, look at the various phylum ranks:

- superphylum == sp
- phylum == p
- subphylum == ps
- infraphylum == pi

I follow the same approach for all other ranks. That is, anything above the standard rank is used as a prefix (e.g. `sp` for `superphylum`), anything l under the standard rank is used as a suffix, e.g. `ps` for `subphylum`.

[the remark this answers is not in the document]
mikerobeson: I think for the various species ranks I'd recommend the following, just to be consistent with the approach I outlined above:
- species group (super species) == ss
- species == s
- species subgroup ==ssg
- subspecies == ssb

mikerobeson:
I ran same query again but with several different ranks specified:

>qiime rescript get-ncbi-data --p-query "txid7898[ORGN] AND (cytochrome c oxidase subunit 1[Title] OR cytochrome c oxidase subunit I[Title] OR cytochrome oxidase subunit 1[Title] OR cytochrome oxidase subunit I[Title] OR COX1[Title] OR CO1[Title] OR COI[Title]) AND mitochondrion[Filter] NOT environmental sample[Title] NOT environmental samples[Title] NOT environmental[Title] NOT uncultured[Title] NOT unclassified[Title] NOT unidentified[Title] NOT unverified[Title]" --o-sequences ~/Desktop/co1-actinopterygii-seqs.qza --o-taxonomy ~/Desktop/co1-actinopterygii-tax.qza --p-ranks domain superkingdom subphylum infraclass suborder family genus --verbose

and received this error:

>Traceback (most recent call last):
  File "/Users/robesonmichael/AnalysisSoftware/miniconda3/envs/qiime2-2020.6-test/lib/python3.6/site-packages/q2cli/commands.py", line 329, in __call__
    results = action(**arguments)
  File "<decorator-gen-561>", line 2, in get_ncbi_data
  File "/Users/robesonmichael/AnalysisSoftware/miniconda3/envs/qiime2-2020.6-test/lib/python3.6/site-packages/qiime2/sdk/action.py", line 245, in bound_callable
    output_types, provenance)
  File "/Users/robesonmichael/AnalysisSoftware/miniconda3/envs/qiime2-2020.6-test/lib/python3.6/site-packages/qiime2/sdk/action.py", line 390, in _callable_executor_
    output_views = self._callable(**view_args)
  File "/Users/robesonmichael/AnalysisSoftware/github.forks/RESCRIPt/rescript/ncbi.py", line 57, in get_ncbi_data
    taxids, ranks, rank_propagation, entrez_delay)
  File "/Users/robesonmichael/AnalysisSoftware/github.forks/RESCRIPt/rescript/ncbi.py", line 160, in get_taxonomies
    taxonomy = OrderedDict([('NCBI_Division', rec['Division'])])
TypeError: string indices must be integers

rescript/ncbi.py
```diff
@@ -84,28 +89,34 @@ def _get(params, ids=None, entrez_delay=0.):
...
     data = []
     while len(data) < expected_num_records:
         params['retstart'] = len(data)
         time.sleep(entrez_delay)
         r = requests.get(efetch, params=params)
-        r.raise_for_status()
+        if r.status_code != requests.codes.ok:
+            content = parse(r.content)
+            content = list(content.values()).pop()
+            warnings.warn('Download did not finish smoothly. The following '
```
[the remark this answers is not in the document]
BenKaehler: This approach seemed better because it tells the user that something went wrong but also allows the user to check which accessions didn't come back.

We could make it better by raising a firmer warning, for instance the sort that shows up even without `--verbose` specified. Alternatively I could prepare a list of accessions that failed to download and raise an error with those in the message. That message could potentially be quite big, though.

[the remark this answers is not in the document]
nbokulich: maybe some kind of log file would be appropriate? I think outputting a log file as text is fine in this case (there is precedent with deblur), unless if you think that failures should cause a fatal error. (it's not clear to me what would cause this failure)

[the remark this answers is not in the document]
BenKaehler: This error occurs when one of your accession ids is wrong or if something goes wrong on the ncbi end.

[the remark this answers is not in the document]
nbokulich: well that's a shame if we can't differentiate the two errors... sounds like a log is our best bet.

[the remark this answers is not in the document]
BenKaehler: I decided to raise an error instead of a warning. I have added a more descriptive error message that will hopefully be useful in most cases.

rescript/tests/test_ncbi.py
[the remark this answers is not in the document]
BenKaehler: I've updated the code to use the abbreviations and added some extra tests, see below.

rescript/tests/test_ncbi.py
```diff
@@ -43,6 +68,23 @@ def test_get_ncbi_data_accession_ids(self):
         taxa = self.taxa.view(DataFrame).to_dict()
         self.assertEqual(acc_tax, taxa)
 
+    def test_get_ncbi_data_rank_propagation_nonstandard_ranks(self):
```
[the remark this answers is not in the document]
BenKaehler: I've added these tests.

BenKaehler:
I think I've fixed most things. I am having trouble reproducing [this error](https://github.com/bokulich-lab/RESCRIPt/pull/52#issuecomment-654414636) but I will keep trying and let you know how I go.

mikerobeson:
Hi @BenKaehler, I use the latest version of your PR and apparently... I am now unable to [replicate the error](https://github.com/bokulich-lab/RESCRIPt/pull/52#issuecomment-654414636) myself. It works! Something else must have happened. Perhaps an internet issue on my end?

PR #54, opened 2020-06-29
fix taxonomy eval

nbokulich:
Fixes a few minor bugs in `evaluate_taxonomy`. These came to light when testing on the UNITE taxonomy — this method was originally validated using greengenes/silva style taxonomies, but testing on UNITE (which has taxonomies like `k__unidentified;p__unidentified;...` a few issues became apparent:
1. redundant taxonomies were not being counted correctly, because the rank depth was counted after casting to a `set`
2. taxonomies without annotation at ANY level were being miscounted (they were counted as classified as the base level, but they are not!)
3. entropy was incorrectly counted if unique counts 

This PR fixes these issues in a few ways:
1. adds some real taxa strings as test data, including UNITE, GG, SILVA, and old-style SILVA. We can add to this list as we test new data sources.
2. tests real data using a few different rank handles (since rank handle trimming can yield empty taxa strings)
3. label counting is done at level 0 (empty string). this level is not plotted, though (as it would be confusing, lead users to think that level 0 is the base level, which it is not).
4. label counting was fixed to not use sets
5. entropy calculations now use a `max_depth` setting to calculate up to and including `max_depth`. Previously max depth was measured on a per-taxon basis, which led to undercounting of empty/truncated taxonomies

### DIFF
--- a/rescript/evaluate.py
+++ b/rescript/evaluate.py
@@ -74,8 +74,9 @@ def _evaluate_taxonomy(taxonomy, rank_handle_regex):
     summary = summarize_taxonomic_depth(
         taxonomy, rank_handle_regex=rank_handle_regex)
     # Measure taxonomic entropy at each level
+    max_depth = summary.index.max()
     entropy = _taxonomic_entropy(
-        taxonomy, rank_handle_regex=rank_handle_regex)
+        taxonomy, rank_handle_regex=rank_handle_regex, max_depth=max_depth)
     entropy = entropy.merge(summary, left_index=True, right_index=True)
     entropy.index.name = 'Level'
     return entropy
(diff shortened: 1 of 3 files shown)


mikerobeson:
This all looks good to me. I do not have a data set to test it on, but I did download the latest UNITE release, imported the taxonomy file, and ran it through `evaluate-taxonomy`. The results appeared sane to me. That is, the number of unique labels increased more steeply towards `species` / `7` . Number of feature terminating was all at `7`.

nbokulich:
thanks @mikerobeson ! If you want to hit some of the edges, you could try using some rank handle regexes (see the tests for an appropriate regex for ignoring UNITE unidentified seqs) to see how that looks — that replicates what I have already done, though, so no worries.

mikerobeson:
Okay, so when I use "^[dkpcofgs]__" or "unidentified" I get similar plots. When I use "^[dkpcofgs]__unidentified" I see changes in the slopes throughout the plot, e.g. many more unclassified features... since I am removing those items completely. If that makes sense?

nbokulich:
yep! perfect! that will some seqs are "unidentified" at each rank, so that causes some to be totally unknown, and others to be truncated if they are not annotation all the way to species level.

PR #58, opened 2020-07-02
refactor dereplicate to use skbio for seq subsetting

### LINKED ISSUE #18: ENH: use DNAIterator instead of pandas for handling large fasta files?
`dereplicate._dereplicate_taxa` uses `pd.Series` as the view type for fasta files. This is convenient for small files but may pose problems with very large fasta files. `DNAIterator` or `DNAFASTAFormat` may be a better way to iterate across sequences.

Other functions might also use pd.Series for handling fasta files.

**Need to test a bit further to see if this shaves off runtime/memory use.**

nbokulich:
fixes #18

replaces pandas with skbio for subsetting large sequence sets. skbio is now used for all sequence processing.

in the process of looking for other places where pandas is used for handing seqs, I found some misleading outdated docstrings in `cross_validate.py` so edited these comments. 

So far it looks like swapping this out did not lead to any real performance improvements (tested dereplicating a large set of sequences) but I did not check memory usage, and the code looks cleaner now without inline artifact imports!

### DIFF
--- a/rescript/cross_validate.py
+++ b/rescript/cross_validate.py
@@ -25,8 +25,8 @@ def evaluate_fit_classifier(ctx,
                             n_jobs=1,
                             confidence=0.7):
     '''
-    taxonomy: pd.Series of taxonomy labels
-    sequences: pd.Series of sequences
+    taxonomy: FeatureData[Taxonomy] artifact of taxonomy labels
+    sequences: FeatureData[Sequence] artifact of sequences
     k: number of kfold cv splits to perform.
     '''
     # Validate inputs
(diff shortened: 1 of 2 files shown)


mikerobeson:
Just in case you do not get see the text I wrote for the PR merge, I'll reiterate here....

I tested the original and new implementation of `dereplicate` on V4 reference sequence data. The original implementation had a memory usage that hovered ~ 880 MB, with a spike up to 1 GB. The new implementation of this PR was much lower, the memory usage hovered ~ 330 MB with a spike up to ~ 500 MB! Awesome! We'll probably get more "bang for our buck" with larger data sets. :-)

nbokulich:
awesome! thanks so much for testing @mikerobeson ! (and for reiterating here)

PR #59, opened 2020-07-05
add super LCA and majority LCA merging

nbokulich:
`merge_taxa` has two new modes:
1. `majority` performs LCA merging, with priority to majority labels
2. `super` performs LCA merging, with priority to majority labels. Labels that are substrings of other labels (including empty rank handles) are collapsed prior to majority/LCA.

In general also includes a decent code refactor and addition of several more tests.

### DIFF
--- a/rescript/_utilities.py
+++ b/rescript/_utilities.py
@@ -10,6 +10,7 @@ from itertools import zip_longest, takewhile
 from re import sub
 import subprocess
 import skbio
+from collections import Counter
 from q2_types.feature_data import DNAFASTAFormat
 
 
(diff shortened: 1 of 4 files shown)


rescript/_utilities.py
```diff
@@ -21,15 +22,29 @@
...
-    '''
-    Find least common ancestor between two semicolon-delimited strings.
-    Input consists of two series containing index "Taxon".
-    Returns the LCA of those taxonomies as an iterator.
-    '''
-    t1 = t1['Taxon']
-    t2 = t2['Taxon']
-    return _find_lca([t1, t2])
+# modified version of _find_lca that prioritizes majority and supersets
+def _find_super_lca(taxa, collapse_substrings=True):
```
mikerobeson, line 26: I'm glad you added this! I had been meaning to inquire about such an approach, as sintax/utax (if I remember correctly) does something similar. That is, going for a more specific taxonomy string when it does not contradict upper-level taxonomy.

rescript/plugin_setup.py
```diff
@@ -215,7 +215,12 @@
                 'taxonomy with the highest score (e.g., confidence or '
                 'consensus score). Note that "score" assumes that this score '
                 'is always contained as the second column in a feature '
-                'taxonomy dataframe.',
+                'taxonomy dataframe. "majority" finds the LCA consensus while '
+                'giving preference to majority labels. "super" finds the LCA '
+                'consensus while giving preference to majority labels and '
+                'collapsing substrings into superstrings (e.g., '
+                '"g__Lactobacillus" would be collapsed with '
+                '"g__Lactobacillus_group_A").',
```
mikerobeson, line 223: I found this description confusing. Perhaps this:

>"super" finds the LCA consensus while giving preference to majority labels and collapsing substrings into superstrings. For example, when a more specific taxonomy does not contradict a less specific taxonomy, the more specific one is chosen, if it is the majority ( e.g. "g__Faecalibacterium; s__prausnitzii", will be preferred over "g__Faecalibacterium; s__" ).

Of course, this is assuming I am following the concept of "super" correctly, given both  the `test_merge_taxa_super_lca` and `test_merge_taxa_super_lca_majority` test cases. Otherwise, I am not sure what this help text is actually telling me.

mikerobeson:
This looks really great @nbokulich! My only concerns are really the help text, of "super". Otherwise everything seems to run well locally, all tests pass.

nbokulich:
Thanks @mikerobeson ! I fixed the description text with your recommendation.

Ultimately, I think usage will be best demonstrated with a detailed tutorial, on the to-do list. For now hopefully it is clear enough to guide users!

We may also want to consider what mode to use as the default for this method. For now I have kept as `len` (choose the taxonomy with the most elements) but `super` sort of has the best of all worlds (hence the name) so we may want to promote that.

mikerobeson:
No problem @nbokulich, I am leaning towards the `super` approach as the default. They way I look at it, we are making use of the most information available. I agree about clarifying details more within a tutorial.

PR #61, opened 2020-07-07
IMP: adds super to dereplicate

mikerobeson:
Adds the same "super"  mode of `merge_taxa` to `dereplicate` for taxonomy dereplication. Help text was updated and two tests cases added.

### DIFF
--- a/rescript/dereplicate.py
+++ b/rescript/dereplicate.py
@@ -13,7 +13,8 @@ import shutil
 
 from q2_types.feature_data import DNAFASTAFormat
 
-from ._utilities import run_command, _find_lca, _majority, _rank_handles
+from ._utilities import (run_command, _find_lca, _majority, _rank_handles,
+                         _find_super_lca)
 
 
 def dereplicate(sequences: DNAFASTAFormat,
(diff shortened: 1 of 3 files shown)


rescript/dereplicate.py
```diff
@@ -139,7 +140,11 @@ def _dereplicate_taxa(taxa, raw_seqs, derep_seqs, uc, mode):
         if mode == 'lca':
             derep_taxa = derep_taxa.apply(lambda x: ';'.join(
                 _find_lca([y.split(';') for y in x]))).to_frame()
-            # find majority taxon within each cluster
+        # find majority superset LCA within each cluster
+        if mode == 'super':
```
nbokulich, line 144: should be `elif`, not that it changes the functionality (since there's only one mode!) but it helps clarify

rescript/plugin_setup.py
```diff
@@ -289,7 +289,13 @@
                 'ancestor among all taxa sharing a sequence. "majority" will '
                 'find the most common taxonomic label associated with that '
                 'sequence; note that in the event of a tie, "majority" will '
-                'pick the winner arbitrarily.',
+                'pick the winner arbitrarily. "super" finds the LCA '
+                'consensus while giving preference to majority labels and '
```
nbokulich, line 293: if any of this text is redundant with the text for `merge_taxa`, I recommend just defining the text string as a variable and appending in both descriptions rather than writing redundant lines of text.

rescript/dereplicate.py
mikerobeson: oops! right!

PR #63, opened 2020-07-16
added filter-taxa method

nbokulich:
allows filtering a `FeatureData[Taxonomy]` artifact based on one or more search criteria.

I decided to add this because, while Q2 style is to typically just subset taxonomy, RESCRIPt evaluate taxonomy methods sometimes require taxonomy to be subset (e.g., if evaluating a specific clade, and some methods do not allow superset taxonomies, e.g.,` evaluate-classifications`)

### DIFF
--- a/rescript/filter_length.py
+++ b/rescript/filter_length.py
@@ -110,12 +110,10 @@ def filter_taxa(taxonomy: pd.Series, ids_to_keep: qiime2.Metadata = None,
         _index_is_superset(set(ids_to_keep), ids)
 
     if include:
-        include = '|'.join(include)
-        ids = ids[taxonomy.str.contains(include.replace(',', '|'))]
+        ids = ids[taxonomy.str.contains('|'.join(include))]
 
     if exclude:
-        exclude = '|'.join(exclude)
-        bad = taxonomy.index[taxonomy.str.contains(exclude.replace(',', '|'))]
+        bad = taxonomy.index[taxonomy.str.contains('|'.join(exclude))]
         ids = ids.difference(bad)
 
     # if not using exclude or include, we only want explicit ids_to_keep
--- a/rescript/filter_length.py
+++ b/rescript/filter_length.py
@@ -6,6 +6,7 @@
 # The full license is in the file LICENSE, distributed with this software.
 # ----------------------------------------------------------------------------
 
+import qiime2
 import pandas as pd
 from q2_types.feature_data import DNAFASTAFormat, DNAIterator
 
@@ -96,6 +97,47 @@ def filter_seqs_length_by_taxon(sequences: DNAFASTAFormat,
     return result, failures
 
 
+def filter_taxa(taxonomy: pd.Series, ids_to_keep: qiime2.Metadata = None,
+                include: str = None, exclude: str = None) -> pd.Series:
+    if include is exclude is ids_to_keep is None:
+        raise ValueError('No filtering criteria were applied!')
+
+    ids = taxonomy.index
+    print('Input features: ' + str(len(ids)))
+    # some initial validation
+    if ids_to_keep:
+        ids_to_keep = ids_to_keep.ids
+        _index_is_superset(set(ids_to_keep), ids)
+
+    if include:
+        include = '|'.join(include)
+        ids = ids[taxonomy.str.contains(include.replace(',', '|'))]
+
+    if exclude:
+        exclude = '|'.join(exclude)
+        bad = taxonomy.index[taxonomy.str.contains(exclude.replace(',', '|'))]
+        ids = ids.difference(bad)
+
+    # if not using exclude or include, we only want explicit ids_to_keep
+    if include is exclude is None:
+        ids = ids_to_keep
+    # otherwise add back ids_to_keep for explicit inclusion after filtering
+    elif ids_to_keep:
+        ids = ids.union(ids_to_keep)
+
+    filtered_ids = len(ids)
+    print('Output features: ' + str(filtered_ids))
+
+    if filtered_ids == 0:
+        raise ValueError("All features were filtered, resulting in an "
+                         "empty collection of taxonomies.")
+
+    taxonomy = taxonomy.reindex(ids)
+    taxonomy.index.name = 'Feature ID'
+
+    return taxonomy
+
+
 def _seq_length_within_range(sequence, taxahits, mins, maxs, global_min,
                              global_max):
     '''
@@ -155,5 +197,5 @@ def _index_is_superset(index1, index2):
     '''
     diff = index1.difference(index2)
     if len(diff) > 0:
-        raise ValueError('The following sequences are missing from '
+        raise ValueError('The following IDs are missing from '
                          'the taxonomy: ' + ', '.join(sorted(diff)))
--- a/rescript/tests/test_filter_length.py
+++ b/rescript/tests/test_filter_length.py
@@ -9,6 +9,7 @@
 
 import pandas as pd
 import qiime2
+import pandas.util.testing as pdt
 from qiime2.plugin.testing import TestPluginBase
 from q2_types.feature_data import DNAIterator
 from qiime2.plugins import rescript
@@ -129,8 +130,7 @@ class TestFilterByTaxonomy(TestPluginBase):
         missing_taxa = import_data(
             'FeatureData[Taxonomy]',
             self.taxa.view(pd.Series).drop(['C1', 'C2']))
-        with self.assertRaisesRegex(
-                ValueError, "sequences are missing.*C1, C2"):
+        with self.assertRaisesRegex(ValueError, "IDs are missing.*C1, C2"):
             rescript.actions.filter_seqs_length_by_taxon(
                 sequences=self.seqs, taxonomy=missing_taxa,
                 labels=['Bacteria'], min_lens=[1200])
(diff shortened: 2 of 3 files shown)


rescript/filter_length.py
```diff
@@ -96,6 +97,47 @@ def filter_seqs_length_by_taxon(sequences: DNAFASTAFormat,
...
+    ids = taxonomy.index
+    print('Input features: ' + str(len(ids)))
+    # some initial validation
+    if ids_to_keep:
+        ids_to_keep = ids_to_keep.ids
+        _index_is_superset(set(ids_to_keep), ids)
+
+    if include:
+        include = '|'.join(include)
+        ids = ids[taxonomy.str.contains(include.replace(',', '|'))]
```
mikerobeson, line 114: Took me a while to figure out what was going on here. I suppose this combination is to protect against whether or not a user is placing spaces and/or commas in between the taxonomies they wish to filter? That is either "Bacteria Archaea" or "Bacteria,Archaea" will work here. At leas that is what seems to be the case playing around in ipython. I like it.

rescript/tests/test_filter_length.py
```diff
@@ -299,3 +299,83 @@ def test_filter_seqs_length_no_failures(self):
...
+        exp_taxa.index.name = 'Feature ID'
+        filtered, = rescript.actions.filter_taxa(
+            self.taxa, ids_to_keep=ids_to_keep, include=['brevis'],
+            exclude=['o__Bacillales'])
+        pdt.assert_series_equal(filtered.view(pd.Series), exp_taxa)
+
+    def test_filter_taxa_fail_all_filtered_out(self):
+        with self.assertRaisesRegex(ValueError, "All features were filtered"):
+            filtered, = rescript.actions.filter_taxa(
+                self.taxa, exclude=['Bacteria'])
```
mikerobeson, line 381: I had no idea that this was even a thing. *I get it... but I don't get it.* Guess I learned a new use for a `with` statement!

mikerobeson:
This is great @nbokulich! Everything seems to work as intended. I ran the following command:

`qiime rescript filter-taxa --i-taxonomy silva-138-ssu-nr99-tax.qza --m-ids-to-keep-file ids_to_keep.txt --p-include Archaea --p-exclude Euckaryota Bacteria --o-filtered-taxonomy silva-138-ssu-nr99-tax-filt.qza`

With the `ids_to_keep.txt` filled with a mix of Bacteria and Eukaryota. The output did indeed only contain the Bacteria and Eukaryota IDs I specified, everything else was Archaea. Cool!

rescript/filter_length.py
nbokulich: oh thanks for catching! silly oversight on my part, actually... I originally made `include` a `Str`, then changed to a `List[Str]`, so the `replace` is extraneous here. I removed that part.

nbokulich:
thanks @mikerobeson ! I fixed the `contains` statement per your review.

devonorourke:
Thanks for this update guys.
Just wondering how I would input a particular list of taxonomic Order ranks with the `--m-ids-to-keep-file ids_to_keep.txt` parameter/file.  

I'm just wondering about I go about filtering things that aren't at rank1. Curious about how Mike has pointed out before where species labels can be wonky, where you get a bunch or rice-named species. 

Say I have a taxonomy file containing this kind of data:
```
Taxon
k__Animalia;p__Arthropoda;c__Insecta;o__Blue;f__Berry;g__Red;s__Red berry
k__Animalia;p__Arthropoda;c__Insecta;o__Blue;f__Berry;g__Granny;s__Granny berry
k__Animalia;p__Arthropoda;c__Insecta;o__Red;f__Apple;g__Red;s__Red delicious
k__Animalia;p__Arthropoda;c__Insecta;o__Red;f__Apple;g__Granny;s__Granny smith
```

I couldn't see anything in the `qiime rescript filter-taxa --help` output that mentioned how I would go about filtering at a particular level, so I'm guessing I need to put somewhere in the metadata file a string of taxa that includes everything _up to_ that level. So, in the above example, if I wanted to filter at the taxonomic Order rank, keeping just those with any `o__Red` label, and discarding anything with `o__Blue` labels, what exactly would I put in the text file?

I'm guessing that I can't pass just "Red" or "Blue", as the `Red` string itself exists in _both_ of those Order-rank Red/Blue groups.

Sorry to be a bother!

nbokulich:
> I'm guessing that I can't pass just "Red" or "Blue", as the Red string
itself exists in *both* of those Order-rank Red/Blue groups.

yep, this would be a big problem if your taxonomy strings don't contain
rank handles... but they do contain rank handles like "o__", so you can
include or exclude labels like "o__Red" or "o__Blue". If you don't have
rank handles, clever mixture of include/exclude parameters can help you
narrow down (e.g., --p-include Red --p-exclude Blue)

> Just wondering how I would input a particular list of taxonomic Order
ranks with the --m-ids-to-keep-file ids_to_keep.txt parameter/file.

use ids_to_keep just for really explicit searches, e.g., if there's a
specific handful of accessions you wish to isolate or add back in. In
general, include and exclude are what you want to use for filtering out
specific clades.

devonorourke:
Thanks -
Now that I think of it, I'm not sure this is going to work exactly for what I was initially trying to accomplish with `qiime rescript evaluate-classifications`. If I wanted to run the analysis just for, say, the arthropod order **Diptera**, I would need to run `qiime rescript filter-taxa` separately for both the _observed_ and _expected_ .qza files, right?
But this doesn't guarantee that the exact same sequences are retained, does it? Wouldn't there be instances where the _expected_ taxonomic rank differs from the _observed_? Presumably this is really unlikely at the order rank, but somehow I feel like what I need to do is:
1. Filter the _observed_ qza file with `--p-include` flag.
2. Find the sequence IDs that each of these sequences correspond to.
3. Grab the _expected_ taxa matching those sequences.
4. Run `qiime rescript evaluate-classifications`.

Maybe I'm totally messing this up though. It's 8pm, so I'm going to hit pause tonight and pick up tomorrow morning. No rush to respond fellas - thanks!

nbokulich:
You're almost there. Key info: observed and expected are linked by feature
IDs. So:

   1. Filter the *expected* qza file with --p-include flag.
   2. Filter the *observed* taxa matching those sequences (using
   filter-taxa with the filtered observed taxa as ids_to_keep).
   3. Run qiime rescript evaluate-classifications.

A bit convoluted, but would be easy to set up a pipeline to loop through a
list of desired taxa, or ALL unique orders and combined these together in
evaluate-classifications for a per-order accuracy visualization.

devonorourke:
Right. In my tired brain last night I had been thinking of "observed" as the actual sequences and "expected" as the output of the cross validation (oops). Thanks for the filtering strategy - I'll give it a go this morning.

I'd be keen to look at these combined results on a per-order level, but I wanted to note that if you went about the trouble of building a tool that could do that, you might also want to consider the number of sequences in each group. In tidybug, for example, there are huge ranges of counts of each specific order. While all insect orders include many thousands of sequences (highs like 500,000 beetle records, or lows like 5,000 dragonflies), other non-insect arthropod orders can be sparsely represented - some have less than 10 sequences per order. I'm not sure you'd want to give much credibility to so few sequences, or, if you did, you'd at least want to know how many sequences the precision/recall metrics were evaluated by. The existing .qzv doesn't have any mention about that.

On one hand, it maybe the user would specify a minimum number of sequences per group before retaining the output. Even if they weren't sure about the exact composition of the database _per rank_ (e.x. per order), they could pass something like `--p-min-records-per-rank` and then know that the visualization produced for each rank had at least that many records.

On the other, you could always loop through and produce every possible group, but somehow modify the plot so that the number of sequences used to generate the figure was prominently displayed.

devonorourke:
Here's a plot I came up with after fuddling a bit in R. I think this is about what you'd want to be able to do within a .qzv file, allowing for the user to toggle between the various taxa. Note that I've incorporated a color metric to see how many sequences were analyzed for a given group - maybe this color isn't at all necessary when you're looking at them individually in the more dynamic plots you produce in view.qiime2.org, but I think having that info is still important.

image

nbokulich:
The sparsity of some groups is part of the reason I set up the filter to be
user-defined — users should manually select specific groups of interest,
and can always evaluate the number of features and other characteristics
separately.

Just because an order is small, however, does not necessarily mean that its
members would perform that much worse than members of large orders — since
ultimately the presence of near-neighbors (e.g., other strains of the same
species) will be more important for supporting good cross-validation
results than distant neighbors.

displaying the number of sequences per group might be possible though — I
will look into it.

devonorourke:
What I wonder about now, after viewing these data, is how the relative decrease in accuracy in some groups might be influenced by the overall diversity of sequences within a given group. For instance, is classification among psocodeans relatively stronger than other groups (at genus/species rank) because there is high overall sequence similarity among the members of this arthropod order? Or is it because there are just a few different taxonomic groups represented in this order? 

I get a sneaking suspicion that you're going to mention something about Shannon's entropy as being important here...

nbokulich:
yep we have a method for that — see evaluate-seqs

you can filter your seqs by taxonomy before inputting to that visualizer to
look at breakdown by group.

devonorourke:
Thanks @nbokulich - I forgot about that tool, which you pointed out to me probably less than 5 days ago. My brain is truly turning to much these days :(
Got a script running now to calculate the entropy for each of the 16 arthropod orders in the plot above and will post in a few hours once it's done running.

devonorourke:
Here are the entropy calcs @nbokulich :
Maybe it makes more sense to group the bars by the kmers rather than the arthropod orders?

image

PR #70, opened 2020-09-11
CI: secret, secret, I've got a secret

thermokarst:


thermokarst:
This PR is ready to merge - not sure what the test errors are about, but they look to me like there is an issue querying NCBI. If it would be helpful, I am happy to cut a PR that _mocks_ the network calls as necessary, that way we aren't subject to NCBI uptime issues.

cc @nbokulich @mikerobeson

nbokulich:
oh weird, I think @BenKaehler addressed these server-side issues in #66 to (among other things) use `mock`... maybe just re-start the build?

thermokarst:
I have retried several times, it keeps failing at the same step.

thermokarst:
Sorry, hit "Enter" too soon on the last comment! I was going to copy over the relevant bit of the error log, but looks like it passed (on the 4th re-run).

I don't have the old log now, unfortunately, but it was some kind of `KeyError`, after attempting to fetch data from NCBI. 🤷

nbokulich:
okay this transient failure sounds unrelated to this actual PR, thanks for persistent + merging.

@BenKaehler maybe you could take a look just to be sure (and/or @thermokarst if you are interested you could see if the test that was failing is mocked in #66 )... #66 (among other things) makes get-ncbi-data more robust to server-side issues like this so I have a hunch you hit an error that @BenKaehler already fixed

PR #78, opened 2020-09-27
add silva 138.1 download to get-silva-data

mikerobeson:
adds the ability to download silva 138.1 files through get-silva-data

### DIFF
--- a/rescript/get_data.py
+++ b/rescript/get_data.py
@@ -58,14 +58,18 @@ def _assemble_silva_data_urls(version, target, download_sequences=True):
         target = 'SSURef_Nr99'
     insert = ref_map[target]
 
-    # now compile URLs
-    # naming inconsitancy, i.e. folder == 138_1, filenames == 138.1.
-    if version == '138.1':
-        base_url = 'https://www.arb-silva.de/fileadmin/silva_databases/'\
-                   'release_138_1/Exports/'
+    # Now compile URLs
+    base_url = 'https://www.arb-silva.de/fileadmin/silva_databases/'\
+               'release_{0}/Exports/'
+    # Check folder / file naming inconsistency.
+    # e.g. folder name for version 138.1 contains the string "138_1",
+    # while the filenames contain the string "138.1". We'll need to modify
+    # the variable `version` for the folder name.
+    # We'll add to this list as we encounter similar issues.
+    if version in ['138.1']:
+        base_url = base_url.format(version.replace('.', '_'))
     else:
-        base_url = 'https://www.arb-silva.de/fileadmin/silva_databases/'\
-                   'release_{0}/Exports/'.format(version)
+        base_url = base_url.format(version)
 
     # construct file urls
     base_url_seqs = base_url + 'SILVA_{0}_{1}_tax_silva.fasta.gz'.format(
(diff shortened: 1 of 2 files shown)


rescript/get_data.py
```diff
@@ -54,16 +54,25 @@ def _assemble_silva_data_urls(version, target, download_sequences=True):
...
+
     # now compile URLs
-    base_url = 'https://www.arb-silva.de/fileadmin/silva_databases/'\
-               'release_{0}/Exports/'.format(version)
+    # naming inconsitancy, i.e. folder == 138_1, filenames == 138.1.
+    if version == '138.1':
+        base_url = 'https://www.arb-silva.de/fileadmin/silva_databases/'\
+                   'release_138_1/Exports/'
+    else:
+        base_url = 'https://www.arb-silva.de/fileadmin/silva_databases/'\
```
misialq, line 67: Can't we rather do something like `release_{0}/Exports/'.format(version.replace('.', '_'))` and skip the if block altogether? (in case there are more minor releases in the future)

mikerobeson: The `if` block exists because, for release 138.1, we can't use the `version` variable for both the folder and file, as we have been able to do. As per my proceeding note in the code:
`# naming inconsistency, i.e. folder == 138_1, filenames == 138.1.`
We'd still need an `if` block somewhere to replace '.' with '_' for the folder name only, not the file name.  
But any thoughts on how to tidy this up would be great!

mikerobeson: Perhaps something like this? We can just add to the list as we encounter other versions with similar folder / filename inconsistencies?
```
version_dir = version # so we do not break the use of `version` for file names later in the code

base_url = 'https://www.arb-silva.de/fileadmin/silva_databases/release_{0}/Exports/'

if version_dir in ['138.1']:
	version_dir.replace('.', '_')
	base_url = base_url.format(version_dir)
else:
	base_url = base_url.format(version_dir)
```

misialq: Oh, I see now what you mean! But the way I suggested would not replace it in the original `version` (replace is not inplace) - that way it would only replace `_` for the folder (without the if block) and otherwise the original version would be used for file names. Or am I still missing something?

mikerobeson: > (replace is not inplace)

 I knew that... honest... 🤦  apparently I need more ☕ ...

Do you think this sufficient:
```
if version in ['138.1']:
    base_url = 'https://www.arb-silva.de/fileadmin/silva_databases/'\
                      'release_{0}/Exports/'.format(version.replace('.', '_'))
else:
    base_url = 'https://www.arb-silva.de/fileadmin/silva_databases/'\
                      'release_{0}/Exports/'.format(version)
```

misialq: Haha, sure, that works! 😄 Although, part of my point was removing the if-else block altogether so that whenever there is a minor version like 138.1 the `.` gets replaced (and otherwise the replace operation has no effect) - unless you think it better for any reason to be very explicit and specify the versions manually like you did here - your call :)

So only keeping something like:
```
base_url = 'https://www.arb-silva.de/fileadmin/silva_databases/'\
                  'release_{0}/Exports/'.format(version.replace('.', '_'))
```

mikerobeson: I totally get your point. But SILVA is not very consistent, and they may very well, decide to use either a '.' or '_' next time around for other versions. I was thinking that the code to obviously reflect that point. But, I guess it is simple enough to change then later, and keep it clean for now. I'll send an update shortly. Thank you for your comments! :-)

misialq: OK, thanks for the explanation and sorry for being a pain ;) There are all the subtleties of those different DBs that I'm still not aware of... - getting there slowly though! (hopefully)

misialq (approved):
LGTM, thanks @mikerobeson !

mikerobeson:
Thanks @misialq! Feel free to squash and merge when you get a chance.

PR #79, opened 2020-09-28
 IMP: parallelise sequence processing

### LINKED ISSUE #39: parallelize *-seqs methods
these all use skbio to iterate over sequences; splitting into chunks should enable processing in parallel.

misialq:
Closes #39.

Following what was already described in the issue discussion, this PR introduces parallel processing in `cull-seqs` and `evaluate-seqs` using joblib's Parallel manager.

### DIFF
--- a/rescript/plugin_setup.py
+++ b/rescript/plugin_setup.py
@@ -586,13 +586,13 @@ RANK_DESCRIPTION = ('List of taxonomic ranks for building a taxonomy from the '
                     "[default: '" +
                     "', '".join(DEFAULT_RANKS) + "']")
 
-_SILVA_VERSIONS = ['128', '132', '138']
+_SILVA_VERSIONS = ['128', '132', '138', '138.1']
 _SILVA_TARGETS = ['SSURef_NR99', 'SSURef', 'LSURef']
 
 version_map, target_map, _ = TypeMap({
     (Str % Choices('128', '132'),
      Str % Choices('SSURef_NR99', 'SSURef', 'LSURef')): Visualization,
-    (Str % Choices('138'), Str % Choices('SSURef_NR99', 'SSURef')):
+    (Str % Choices('138', '138.1'), Str % Choices('SSURef_NR99', 'SSURef')):
         Visualization,
 })
 
--- a/rescript/plugin_setup.py
+++ b/rescript/plugin_setup.py
@@ -386,26 +386,30 @@ plugin.methods.register_function(
     function=cull_seqs,
     inputs={
         'sequences': FeatureData[Sequence | RNASequence]
-        },
+    },
     parameters={
         'num_degenerates': Int % Range(1, None),
-        'homopolymer_length': Int % Range(2, None)
-        },
+        'homopolymer_length': Int % Range(2, None),
+        'n_jobs': Int % Range(1, None)
+    },
     outputs=[('clean_sequences', FeatureData[Sequence])],
     input_descriptions={
         'sequences': 'DNA or RNA Sequences to be screened for removal based '
                      'on degenerate base and homopolymer screening criteria.'
-        },
+    },
     parameter_descriptions={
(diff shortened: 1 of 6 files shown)


rescript/screenseq.py
```diff
@@ -40,7 +40,11 @@ def cull_seqs(sequences: DNAIterator, num_degenerates: int = 5,
 
     n_jobs = cpu_count()
```
mikerobeson, line 41: We'll want to include `n_jobs` as a user defined value, with a default of 1, and make it accessible through `plugin_setup.y`, [here](https://github.com/bokulich-lab/RESCRIPt/blob/07ceb370663999a9c6c7e279c8a5855477baa33c/rescript/plugin_setup.py#L386).

misialq: Added!

rescript/screenseq.py
```diff
@@ -35,10 +34,9 @@ def _filter_seq(seq_obj, num_degenerates, homopolymer_length):
 
 
 def cull_seqs(sequences: DNAIterator, num_degenerates: int = 5,
-              homopolymer_length: int = 8) -> DNAFASTAFormat:
+              homopolymer_length: int = 8, n_jobs: int = 1) -> DNAFASTAFormat:
     result = DNAFASTAFormat()
 
-    n_jobs = cpu_count()
     parallel = Parallel(n_jobs=n_jobs, backend='loky')
     seqs = parallel(
```
mikerobeson, line 41: I have a naive question. I am not all that familiar with how `Parallel` works... Does this keep all `seq` data in memory prior to writing it out later? If so, we should avoid this. Users often run into memory issues, which is why we try to read/write the the data as it is being processed. Can this be adjusted so that if using 1 CPU, that it will read and write the sequences as it is culling? Or something similar? I think @BenKaehler, has been able to write out separate chunked files using while using multiple jobs, then merged those files together at the end.  At least that is what I think was done. Perhaps I am mis-understanding...

misialq: That's a very good point and yes, that's how it works - everything is stored in memory. I tried before some other approaches but everything was just too slow. I also tried out your suggestion - it works, in principle, but the fact of constantly opening and writing to new files slows it down significantly... (if I do it on a per-sequence basis). As a last approach, I tried writing to the final file directly from the worker processes using a lock - that works pretty nicely and is as fast as the version before. The only drawback here is that the order of sequences in the new file won't necessarily be preserved - do you think this is a problem? In any case, feel free to check out the update in the new commits. Thanks!

mikerobeson: Thank you so much @misialq! I do not think order will be a problem for the intended use cases. We'll be culling sequences anyway. If it will be a problem for some users, then we can simply make note of this in the user help text. If they would like to keep it in order, then they can use 1 CPU. Or... we could do something insane and provide access to both approaches as two different actions. One that is for low-memory users, and another as you initially had it for those with ample memory. Though that might make things more complicated... We can always return to this if it is a problem.  I'll take a look at the new commit now. :-)

rescript/plugin_setup.py
```diff
@@ -386,26 +386,30 @@
...
         'sequences': 'DNA or RNA Sequences to be screened for removal based '
                      'on degenerate base and homopolymer screening criteria.'
-        },
+    },
     parameter_descriptions={
         'num_degenerates': 'Sequences with N, or more, degenerate bases will '
                            'be removed.',
         'homopolymer_length': 'Sequences containing a homopolymer sequence of '
                               'length N, or greater, will be removed.',
+        'n_jobs': 'Number of concurrent processes to use while processing '
```
mikerobeson, line 405: Perhaps add something like this: "Output sequence order may change when using multiple jobs."

misialq: Thanks @mikerobeson! - added and merging.

PR #87, opened 2020-10-28
IMP: add fasta subsampling action

misialq:
This PR adds a `subsample-fasta` action that can be used to randomly sample from FASTA (either plain or aligned DNA). User can define either a fraction of the original sequence count or a number of sequences they want the sample to have.

Some open question/comments:
1. I tried to use the TypeMatch to make it work with both, plain and aligned, DNA sequences but I seem to be unable to make it work when it comes to setting the correct types in the `subsample_fasta` function definition (tried with Union, doesn't work) - hoping for some hints here :)
2. this implementation is a bit different from the [suggested solution](https://github.com/biocore/qiime/blob/76d633c0389671e93febbe1338b5ded658eba31f/qiime/util.py#L1536) so that:
- it is possible to specify an absolute number of sequences to sample
- the sample always contains the exact number of sequences (with the above approach it is not guaranteed)

### DIFF
--- a/rescript/_utilities.py
+++ b/rescript/_utilities.py
@@ -107,6 +107,10 @@ def _read_dna_fasta(path):
     return skbio.read(path, format='fasta', constructor=skbio.DNA)
 
 
+def _read_dna_alignment_fasta(path):
+    return skbio.TabularMSA.read(path, format='fasta', constructor=skbio.DNA)
+
+
 def _rna_to_dna(path):
     ff = DNAFASTAFormat()
     with ff.open() as outfasta:
--- a/rescript/plugin_setup.py
+++ b/rescript/plugin_setup.py
@@ -8,9 +8,11 @@
 
 import importlib
 
+from qiime2.core.type import TypeMatch
 from qiime2.plugin import (Str, Plugin, Choices, List, Citations, Range, Int,
                            Float, Visualization, Bool, TypeMap, Metadata)
 
+from .subsample import subsample_fasta
 from .trim_alignment import trim_alignment
 from .merge import merge_taxa
 from .dereplicate import dereplicate
@@ -826,6 +828,34 @@ plugin.pipelines.register_function(
         "Subsequently, start (5'-most) and end (3'-most) position from fwd "
         "and rev primer located within the new alignment is identified and "
         "used for slicing the original alignment."),
+)
+
+T = TypeMatch([AlignedSequence])
+plugin.methods.register_function(
+    function=subsample_fasta,
+    inputs={'sequences': FeatureData[T], },
+    parameters={
+        'subsample_size':
+            Float % Range(0, 1, inclusive_start=False, inclusive_end=True) |
+            Int % Range(1, None),
+        'random_seed': Int % Range(1, None)
+    },
+    outputs=[('sample_sequences', FeatureData[T]), ],
+    input_descriptions={'sequences': 'Sequences to subsample from.', },
+    parameter_descriptions={
+        'subsample_size': 'Size of the sample: if a number 0.0-1.0 is given '
+                          'a corresponding fraction of sequences will be '
+                          'sampled. If an integer >1 is given, a corresponding'
+                          'number of sequences will be sampled.',
+        'random_seed': 'Seed to be used for random sampling.'
+    },
+    output_descriptions={
+        'sample_sequences': 'Sample of original sequences.', },
+    name='Subsample an indicated number of sequences from a FASTA file.',
+    description=(
+        "Subsample a set of sequences (either plain or aligned DNA)"
+        "based on the number of desired sequences or a fraction of "
+        "original sequences."),
 )
 
 # Registrations
--- a/rescript/subsample.py
+++ b/rescript/subsample.py
@@ -1,0 +1,49 @@
+import random
+from typing import Union
+
+from q2_types.feature_data import (AlignedDNAFASTAFormat, DNAFASTAFormat)
+
+from rescript._utilities import _read_dna_fasta, _read_dna_alignment_fasta
+
+
+def _read_seqs(sequences: Union[DNAFASTAFormat, AlignedDNAFASTAFormat]):
+    return _read_dna_fasta(str(sequences)) if isinstance(
+        sequences, DNAFASTAFormat) else _read_dna_alignment_fasta(
+        str(sequences))
+
+
+def subsample_fasta(sequences: AlignedDNAFASTAFormat,
+                    subsample_size: Union[float, int] = 0.1,
+                    random_seed: int = 42) -> \
+        AlignedDNAFASTAFormat:
+    # set seed for reproducibility
+    random.seed(random_seed)
+
+    # get all ids
+    seqs_iter = _read_seqs(sequences)
+    ids = [seq.metadata['id'] for seq in seqs_iter]
+
+    # get a sample of the ids
+    if isinstance(subsample_size, float):
+        ids_sample = set(random.sample(ids, int(subsample_size * len(ids))))
+    else:
+        if subsample_size <= len(ids):
+            ids_sample = set(random.sample(ids, subsample_size))
+        else:
+            raise ValueError(f"The requested sample size {subsample_size} "
+                             f"is bigger than the total number of "
+                             f"sequences {len(ids)}")
+
+    # write out sampled sequences
+    result = DNAFASTAFormat() if isinstance(
+        sequences, DNAFASTAFormat) else AlignedDNAFASTAFormat()
+
+    seqs_iter = _read_seqs(sequences)
+    with result.open() as out_fasta:
+        for seq in seqs_iter:
+            id = seq.metadata['id']
+            if id in ids_sample:
+                seq.write(out_fasta)
+                ids_sample.remove(id)
+
+    return result


rescript/plugin_setup.py
```diff
@@ -828,6 +830,34 @@
         "used for slicing the original alignment."),
 )
 
+T = TypeMatch([AlignedSequence])
```
mikerobeson, line 833: Quick comment on `TypeMatch`. I am in the process of figuring this out myself. I think that the [code here](https://github.com/qiime2/q2-feature-table/blob/9ed6160adad45445ec054e5ce034a3b3ba25a9b4/q2_feature_table/plugin_setup.py#L379), is a great example of what we'd want. This is part of the `qiime feature-table filter-seqs` command. Apologies, if you've already tried this.

misialq: I've looked through many examples but did not see that one - thanks! I think it would work if we could stick with the `DNAIterator` as input type, however I did not want to use it as I need to iterate over the sequences twice for it to work the way I described above. That's why I'm creating the iterator "manually" within  `_read_seqs`... :/ 

You gave me another idea though - will try tomorrow!

mikerobeson: You just headed me off at the pass! I was just going to ask why you were not using DNAIterator. :-)

mikerobeson: I think you can just simply just set `seek(0)` after running through the first time with DNAIterator.

misialq: OK, I think I got it to work (thanks for the hints @nbokulich and @mikerobeson!) - you be the judge if that makes sense. 

I don't think using `seek` works here - at least it didn't when I tried - am I missing something? Instead, I used the `tee` functionality from itertools to make a clone of the original iterator and that worked too. I ended up using `DNAIterator` as both, input and output, and created a transformer `DNAIterator` -> `AlignedDNAFASTAFormat` to allow converting back to appropriate output type. Let me know what you think. Thanks!

Tested with both input types on a file with 1mln 400bp-long sequences - worked as expected.

mikerobeson: This looks awesome @misialq! I'll take it for a spin now.

rescript/_utilities.py
```diff
@@ -107,9 +107,19 @@ def _read_dna_fasta(path):
     return skbio.read(path, format='fasta', constructor=skbio.DNA)
 
 
+def _read_dna_alignment_fasta(path):
+    return skbio.TabularMSA.read(path, format='fasta', constructor=skbio.DNA)
```
mikerobeson, line 111: I think we should avoid using `skbio.TabularMSA`. This basically reads the entire alignment into memory, which is what we are trying to avoid. I had to kill the process on my laptop.

mikerobeson: Oh, I see, this is just for the test cases. So we should be are okay there.

rescript/_utilities.py
```diff
@@ -107,9 +107,19 @@ def _read_dna_fasta(path):
...
     with ff.open() as outfasta:
         for seq in _read_rna_fasta(path):
             seq.reverse_transcribe().write(outfasta)
     return ff
+
+
+def _dna_iterator_to_aligned_fasta(iterator):
+    ff = AlignedDNAFASTAFormat()
+    skbio.io.write(iter(iterator), format='fasta', into=str(ff))
+    return ff
```
mikerobeson, line 125: Why is this not structured similar to `_rna_to_dna` above? or vice-versa? Is there something I am missing?

I am not sure if it is this step, or another, but this PR is now using more memory compared to  prior implementation. 🤷‍♂️   I am testing this out on the full NR99 SILVA alignment.

misialq: Hmm, that's odd. I'll try to dig a bit more to see what the cause is.

About that piece of code: that's how this kind of conversion was done in q2-types so I borrowed it from there. Can try to re-write to resemble what you mentioned and see whether that had any impact.

rescript/subsample.py
```diff
@@ -0,0 +1,47 @@
...
+                             f"sequences {len(ids)}")
+
+    # create new iterator that will yield sequences
+    # based on the list generated above
+    def sample_from_iterator():
+        for seq in seqs_iter_clone:
+            id = seq.metadata['id']
+            if id in ids_sample:
+                ids_sample.remove(id)
+                yield seq
```
mikerobeson, line 45: Doesn't this yield all of the sequences into memory? Shouldn't we write to the iterator like [this](https://github.com/bokulich-lab/RESCRIPt/blob/20492bd8bcb676fe552d0085938587bdfc34cd1e/rescript/degap.py#L12)?

rescript/_utilities.py
nbokulich: I have seen examples in the skbio cookbooks like this... I think the structure is fine, but @misialq if you do a quick benchmark it would be great to know 😄

rescript/subsample.py
```diff
@@ -0,0 +1,47 @@
...
+
+import random
+from itertools import tee
+from typing import Union
+from q2_types.feature_data import DNAIterator
+
+
+def subsample_fasta(sequences: DNAIterator,
+                    subsample_size: Union[float, int] = 0.1,
+                    random_seed: int = 42) -> DNAIterator:
```
nbokulich, line 17: just set random seed default to 1... otherwise you will get questions from people who think there's some magic going on here!

rescript/subsample.py
```diff
@@ -0,0 +1,47 @@
...
+
+    # get all ids
+    ids = [seq.metadata['id'] for seq in seqs_iter]
+
+    # get a sample of the ids
+    if isinstance(subsample_size, float):
+        ids_sample = set(random.sample(ids, int(subsample_size * len(ids))))
+    else:
+        if subsample_size <= len(ids):
+            ids_sample = set(random.sample(ids, subsample_size))
```
nbokulich, line 32: this if/else is getting complex, and there's a bit of code repetition. I recommend using the if/else to redefine sample_size, then sample ids outside of the if/else.

How about this?

    n_seqs = len(ids)
    if isinstance(subsample_size, float):
        subsample_size = int(subsample_size * n_seqs)
    elif subsample_size > n_seqs:
        raise ValueError(f"The requested sample size {subsample_size} "
                             f"is bigger than the total number of "
                             f"sequences {n_seqs}")
ids_sample = set(random.sample(ids, subsample_size))

rescript/subsample.py
```diff
@@ -0,0 +1,47 @@
...
+                             f"is bigger than the total number of "
+                             f"sequences {len(ids)}")
+
+    # create new iterator that will yield sequences
+    # based on the list generated above
+    def sample_from_iterator():
+        for seq in seqs_iter_clone:
+            id = seq.metadata['id']
+            if id in ids_sample:
+                ids_sample.remove(id)
```
nbokulich, line 44: ids should be unique, so I don't think think there's any need to remove from the list of ids

rescript/plugin_setup.py
```diff
@@ -828,6 +830,34 @@
         "used for slicing the original alignment."),
 )
 
+T = TypeMatch([AlignedSequence, Sequence])
+plugin.methods.register_function(
+    function=subsample_fasta,
+    inputs={'sequences': FeatureData[T], },
```
nbokulich, line 836: no need for the trailing comma

rescript/plugin_setup.py
```diff
@@ -828,6 +830,34 @@
...
+    parameters={
+        'subsample_size':
+            Float % Range(0, 1, inclusive_start=False, inclusive_end=True) |
+            Int % Range(1, None),
+        'random_seed': Int % Range(1, None)
+    },
+    outputs=[('sample_sequences', FeatureData[T]), ],
+    input_descriptions={'sequences': 'Sequences to subsample from.', },
+    parameter_descriptions={
+        'subsample_size': 'Size of the sample: if a number 0.0-1.0 is given '
```
nbokulich, line 846: I recommend changing the description to:

"Size of the random sample. Either a fraction of the total count (if a float) or a positive number (if an integer)."

No need to list the ranges in the description, as this information will be reported automatically.

nbokulich (changes requested):
This is looking great @misialq ! I have a few small comments (I intended to just leave one or two small comments, then wound up doing most of a review... so sorry the comments are scattered inline)

misialq:
OK. First of all, I found the reason for the enormous memory consumption - it was the cloned iterators. Supposedly, when one doesn't iterate over the two of them more-or-less in parallel (what was the case here), the entire contents of one of them is effectively loaded into memory 🤦  
So, that's not the way to go. Since that was the only way to make it work with support for specifying exact number of sequences (see original PR description, point 2), after a discussion with @nbokulich I concluded we'll drop that support for now in favour of a simpler solution suggested initially by @mikerobeson that works with both, DNA and AlignedDNA, types of sequences.

Thanks for your reviews and suggestions, @mikerobeson and @nbokulich !

rescript/subsample.py
misialq: I don't think it does - this generator would still load element by element and only return (or yield rather) when the `id` was in the list of ids.

rescript/subsample.py
misialq: Since for large lists (or sets) it can be costly to every time look up if an element is present, the idea was to shrink that set at every iteration as to make the lookups progressively faster. In any case, I guess it does not matter anymore 😭

rescript/plugin_setup.py
misialq: But it `flake`s!

misialq: Oh, wait, it actually doesn't - sorry 🤦  Will clean up!

mikerobeson:
Wow that was some great detective work @misialq! I'll take it out for a spin today.

misialq:
Hope it doesn't crash your laptop this time! 🤞 😅

rescript/subsample.py
```diff
@@ -0,0 +1,25 @@
...
+# The full license is in the file LICENSE, distributed with this software.
+# ----------------------------------------------------------------------------
+
+import random
+from q2_types.feature_data import DNAIterator
+
+
+def subsample_fasta(sequences: DNAIterator,
+                    subsample_size: float = 0.1,
+                    random_seed: int = 42) -> DNAIterator:
```
mikerobeson, line 15: Probably a good idea to set the seed to either a typical value like `1`, `123`, or `999`, or something. Although I personally like `42` the answer to the meaning of life, the universe, and everything... it'll be lost on most. :-)

misialq: Oooops, sorry for that! I had it changed already but then must have CMD+Zed too far 😅  - will change again!

mikerobeson (approved):
This all looks great to me.

PR #91, opened 2021-01-04
ENH: use add-fragments flag when aligning primers for trimming

misialq:
Adds the `addfragments` support (introduced in q2-alignment: https://github.com/qiime2/q2-alignment/pull/71/) when aligning primers before trimming sequences. 

Updates the year info in the licensing part of all files.

### DIFF
--- a/rescript/tests/test_degap.py
+++ b/rescript/tests/test_degap.py
@@ -25,8 +25,7 @@ class TestDegapSeq(TestPluginBase):
         super().setUp()
         input_fp = self.get_data_path('degap-test-alignment.fasta')
         self.alignedseqs = AlignedDNAFASTAFormat(
-            input_fp, mode='r').view(
-            AlignedDNAIterator)
+            input_fp, mode='r').view(AlignedDNAIterator)
 
     def test_degap_seqs(self):
         #  remove all '-' and '.' chars.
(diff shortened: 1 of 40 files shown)


rescript/tests/test_degap.py
```diff
@@ -25,8 +25,8 @@ def setUp(self):
         super().setUp()
         input_fp = self.get_data_path('degap-test-alignment.fasta')
         self.alignedseqs = AlignedDNAFASTAFormat(
-                                input_fp, mode='r').view(
-                                AlignedDNAIterator)
+            input_fp, mode='r').view(
```
nbokulich, line 28: minor: place on one line?

rescript/trim_alignment.py
```diff
@@ -220,7 +220,9 @@ def _trim_alignment(expand_alignment_action,
 
         # expand the existing alignment by addition of primers
         alignment_with_primers = expand_alignment_action(
```
nbokulich, line 222: the more q2-canonical way to write this would be:
```
         alignment_with_primers, = expand_alignment_action(
            alignment=aligned_sequences,
            sequences=primers,
            addfragments=True)
```

nbokulich (changes requested):
Thanks @misialq ! just a couple very minor comments inline

also a general question: we always want to use add-fragments which is why this is hardcoded, right? Just asking because you say "support" so I am double-checking that this is not intended to be exposed... from this discussion (https://github.com/bokulich-lab/RESCRIPt/pull/85#issuecomment-713042575 and below) this sounds like the case but just want to be sure I understand:

misialq:
Thanks for review, @nbokulich. Yes, we always want to use `addfragments` as it is optimised for adding short sequences to existing alignments. Sorry for the confusing wording.

nbokulich (approved):
looks great, thanks @misialq !

PR #100, opened 2021-03-22
issue-98 : updated minimum qiime 2 version in docs

### LINKED ISSUE #98: update documentation for 2021.2.0
the installation docs etc are a bit out of date, e.g, regarding minimum versions etc.

For example, [see this forum issue](https://forum.qiime2.org/t/rescript-error-default-value-for-parameter-reads-per-batch-is-not-of-semantic-qiime-type-int-range-0-none-or-none/18792?u=nicholas_bokulich).

mikerobeson:
Sets minimum QIIME 2 version to `2021.2`. Fixes #98 

Is there anything else that we should add / change within docs of this PR?

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,10 +32,10 @@ pip install git+https://github.com/bokulich-lab/RESCRIPt.git
 ```
 
 ### Option 2: Install within QIIME 2 environment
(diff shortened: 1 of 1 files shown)


misialq (approved):
I can't think of anything else...

nbokulich:
LGTM thanks @mikerobeson !

PR #101, opened 2021-04-08
A couple of tweaks for get-ncbi-data

BenKaehler:
Fixed `get-ncbi-data` so that it now accommodates the oddness evident in the download that resulted from [this](https://forum.qiime2.org/t/rescript-error-download-did-not-finish-reason-unknown/18928) query.

Added a unit test.

Unfortunately NCBI has fixed a couple of things that were broken before, so I had to loosen one unit test.


mikerobeson:
I just installed an ran this PR, using the query specified in the original forum post. It appears to work... with all the warning message glory! :-)

mikerobeson (approved):
I just installed an ran this PR, using the query specified in the original forum post. It appears to work... with all the warning message glory! The code LGTM! :-)

BenKaehler:
The ci errors on this should resolve after #102 goes in. Thanks @mikerobeson!

misialq:
Hey @BenKaehler, I'm about to release version 2021.4: do you think you could update this PR with the latest changes to fix the CI before I do that? We could also merge it after - I just thought that it all works and it's only a matter of bringing the PR up-to-date.

Issue #105, opened 2021-04-09
Migrate transformers, etc... to qiime2/q2-types


mikerobeson:
Given the recent updates to q2-types (qiime2/q2-types#256) to handle RNA data, and the current required functionality for RESCRIPt (*e.g.* DNA-to-RNA conversions), we should migrate some transformers and related content to `q2-types`. We should slate this for the next QIIME 2 (2021.6??) and RESCRIPt releases.

thermokarst:
This actually makes it so that it is impossible to have both q2-types and rescript installed in the same env:

```bash
ValueError: Duplicate semantic type ('RNASequence') defined in plugins: 'types' and 'rescript'
```

Unfortunately, this has caused some problems with the QIIME 2 2021.4 release - we need to train new feature classifiers in order to publish the release, but we can't train right now because rescript is uninstallable in this state. I'll take a crack at fixing this, PR coming in soon.

thermokarst:
PR: https://github.com/bokulich-lab/RESCRIPt/pull/106

nbokulich:
@thermokarst @mikerobeson can this issue be closed? It looks like this was resolved in #102 ... is there still a lingering transformer that needs to migrate?

mikerobeson:
There are [three transformers](https://github.com/bokulich-lab/RESCRIPt/blob/e73ade00c1a0851eba419499361454405198ee2f/rescript/types/_transformer.py#L61) that need to be moved over still...

nbokulich:
the first utility you linked (`reverse_transcribe`) is exposing the transformer as a stand-alone method and should not be moved (at least not to q2-types).

PR #102, opened 2021-04-08
removal of RESCRIPt RNA types, etc..

mikerobeson:
As RNA types, formats, and transformers have been added to base QIIME 2 as of PR (qiime2/q2-types#256), this created problems with RESCRIPt's own RNA types. This PR removes all RESCRIPt RNA types and uses the new base QIIME 2 RNA types. Some transformers (*i.e.* RNA -> DNA) in RESCRIPt that should be migrated over to `q2-types`. But I left them alone for now.

All unit tests and flake8 checks pass. I've also installed these PR changes into my `qiime2-dev` environment w/o warnings of duplicated types, etc...

PR #106, opened 2021-04-21
maint: remove duplicate types/fmts

thermokarst:
related to (but does not close) #105 - this changeset removes code that is now housed over in q2-types


thermokarst:
I missed #102 - oops!

PR #113, opened 2021-07-13
BUG: fix 'Taxon' column to always be first after merge

### LINKED ISSUE #112: BUG: merge_taxa reorders columns during merge
It seems that when combining DataFrames containing more columns than just 'Taxa' , the columns sometimes get reordered and the 'Taxa' column ends up not being first.

See here for more details: [QIIME2 forum post](https://forum.qiime2.org/t/merging-taxonomies-made-with-different-classifier-types-using-rescripts-merge-taxa/20110)

misialq:
When the DataFrames get _combined_ the columns will be sorted alphabetically, with the preference to first sort the capitalized column names and only then the other ones. That causes the column 'Taxa' to loose its position as a first column (required) when the other column names were provided as capitalized words.

Changes introduced in this PR allow to first capitalize all the column names to ensure consistent sorting behaviour during merge and then ensure that the 'Taxon' column is always first, regardless of the other columns. The tests were adjusted to reflect this behaviour.

Closes #112.

### DIFF
--- a/rescript/_utilities.py
+++ b/rescript/_utilities.py
@@ -72,7 +72,7 @@ def _rank_length(t1, t2):
 
 
 def _find_top_score(t1, t2):
-    return t1 if t1['score'] > t2['score'] else t2
+    return t1 if t1['Score'] > t2['Score'] else t2
 
 
 def _taxon_to_list(taxon, rank_handle):
--- a/rescript/merge.py
+++ b/rescript/merge.py
@@ -23,10 +23,13 @@ def merge_taxa(data: pd.DataFrame,
                mode: str = 'len',
                rank_handle_regex: str = '^[dkpcofgs]__',
                new_rank_handle: str = None) -> pd.DataFrame:
-    # Convert taxonomies to list; optionally remove rank handle
     for d in data:
+        # Convert taxonomies to list; optionally remove rank handle
         d['Taxon'] = d['Taxon'].apply(
             lambda x: _taxon_to_list(x, rank_handle=rank_handle_regex))
+
+        # Capitalize column names for sorting consistency
+        d.columns = [x.capitalize() for x in d.columns]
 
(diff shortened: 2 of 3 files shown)


nbokulich:
Thanks @misialq ! Why not just reorder so that `Taxon` is first, and skip the capitalization? Capitalizing is not a big change, but just seems unnecessary if you are still reordering.

misialq:
That's true - it wouldn't be necessary here. My thinking was to just clean it up to prevent a similar issue from occurring somewhere down the line (not that I can think of any specific place). I can remove that part though if you think that's unlikely.

nbokulich:
I guess the one possible future issue that it addresses is inconsistent capitalization (e.g., merging one df with `consensus` and another with `Consensus`). But this is probably low priority 🤷 as it should not occur now (theoretically it could if importing a taxonomy from elsewhere)

@mikerobeson what's your take on this? Is it more important to preserve original case or more important to prevent possible future bugs regarding capitalization inconsistency? I think I am happy either way, but I always just get a little bit nervous when altering original column names 😬

mikerobeson:
@nbokulich, I am also on the fence about whether or not we should capitalize the column names. However, I would not be surprised, when merging taxonomies from different databases, that a something like this may crop up unexpectedly. I agree that an issue would more likely crop up form a user importing their own files, rather than anything we may parse via RESCRIPt. I'd say keep the capitalization in, as long as it does not require us to update all the test cases.

nbokulich:
Thanks for the input @mikerobeson ! @misialq has already updated the unit tests to capitalize.

My gut feeling is that we should not alter the column names, but I see possible issues in either direction... @misialq I leave the final decision up to you 😁

misialq:
Thanks @both, in that case I think I'll keep the capitalization and pray that the god of dataframes looks graciously upon this change.

PR #114, opened 2021-07-13
BUG: set unresolved taxonomies to 'Unassigned'

### LINKED ISSUE #111: BUG: merge_taxa LCA modes leave empty labels when there is no LCA
The LCA methods will leave blank taxonomies if there is no LCA (e.g., because one is "Unclassified"). 

Should instead fill in a placeholder label (e.g., 'Unassigned'), as blank labels can cause some downstream methods to break. This placeholder label string should be exposed as a parameter so that users can specify placeholder labels that match their pre-existing taxonomies.

Reported [on the Q2 forum](https://forum.qiime2.org/t/rescript-merge-taxa-non-urgent-bug-blank-taxon-values-created-when-using-super/20115?u=nicholas_bokulich).

misialq:
This PR introduces a fix to replace empty taxonomy strings (resulting from unresolved taxonomies) with 'Unassigned' label to prevent downstream issues with the resulting taxonomy (e.g. the pd.DataFrame -> TSVTaxonomyFormat transformer will turn those into NaN values)  

Closes #111.

### DIFF
--- a/rescript/merge.py
+++ b/rescript/merge.py
@@ -71,6 +71,9 @@ def merge_taxa(data: pd.DataFrame,
     else:
         result['Taxon'] = result['Taxon'].apply(lambda x: ';'.join(x))
 
+    # fill unassigned taxa, if any
+    result['Taxon'].replace('', 'Unassigned', inplace=True)
+
     # gotta please the type validator gods
     result.index.name = 'Feature ID'
 
--- a/rescript/merge.py
+++ b/rescript/merge.py
@@ -22,7 +22,8 @@ MODE_ERROR_SCORE = (
 def merge_taxa(data: pd.DataFrame,
                mode: str = 'len',
                rank_handle_regex: str = '^[dkpcofgs]__',
-               new_rank_handle: str = None) -> pd.DataFrame:
+               new_rank_handle: str = None,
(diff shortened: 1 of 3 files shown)


rescript/merge.py
```diff
@@ -71,6 +71,9 @@ def merge_taxa(data: pd.DataFrame,
     else:
         result['Taxon'] = result['Taxon'].apply(lambda x: ';'.join(x))
 
+    # fill unassigned taxa, if any
+    result['Taxon'].replace('', 'Unassigned', inplace=True)
```
nbokulich, line 75: I think we should default to 'Unassigned' but this label should be exposed, so that users can make it consistent with whatever their taxonomy is (e.g., if importing an unusual taxonomy or using taxonomy classifications from something other than q2-feature-classifier). In 99% of cases in the wild 'Unassigned' is fine, but might as well expose. 🤷‍♂️

nbokulich (changes requested):
Thanks @misialq ! minor comment in-line, otherwise I think this is ready to merge once tests pass...

rescript/merge.py
misialq: Gotcha, will add!

nbokulich (approved):
LGTM thanks @misialq !

PR #116, opened 2021-07-21
ENH: replace-taxonomy action

mikerobeson:
This DRAFT PR intends to provide a very simple way to find and replace taxonomy strings. For example, a user may like to fix the `Escherichia-Shigella` garbage, by providing the following information below as a metadata file:

| id | replacements |
| -  |  - |
| g__Salmonella | g__Escherichia-Shigella |
| g__Escherichia; |	g__Escherichia-Shigella; |
| g__Shigella	| g__Escherichia-Shigella |

*Note: although the current implementation uses `re`, it does so with `re.escape`. So, only literal find and replace is carried out. Just keeping it simple for now.*

Must do:
 - test code

Should eventually do:
 - add option to provide a list of pairs on the command line?

Other thoughts?


rescript/plugin_setup.py
```diff
@@ -464,6 +465,46 @@
 )
 
 
+plugin.methods.register_function(
+    function=replace_taxonomy,
+    inputs={
+        'taxonomy': FeatureData[Taxonomy]
+    },
+    parameters={
+        'taxonomy_replacement_map': Metadata
```
nbokulich, line 474: make this a metadata column instead, then it will necessarily be a series (can skip the column count validation)

rescript/replace_taxonomy.py
```diff
@@ -0,0 +1,45 @@
...
+# ----------------------------------------------------------------------------
+
+from qiime2.plugin import Metadata
+import pandas as pd
+import re
+
+
+def make_substitutions_dict(taxonomy_replacement_map):
+    if taxonomy_replacement_map.column_count < 1:
+        raise ValueError("Metadata file should only contain one column of "
```
nbokulich, line 16: can skip this validation if you accept a metadata column instead of metadata

rescript/plugin_setup.py
mikerobeson: Whoa cool! Done!

rescript/replace_taxonomy.py
mikerobeson: Super duper cool! Done.

rescript/plugin_setup.py
```diff
@@ -464,6 +466,46 @@
...
+                                    '\'g__Escherichia;\' with '
+                                    '\'g__Escherichia-Shigella;\'. All items '
+                                    'in the first column must be unique!'},
+    output_descriptions={
+        'edited_taxonomy': 'Taxonomy in which the original rank-lineage '
+                           'strings were replaced by user-supplied '
+                           'rank-lineage strings.'
+    },
+    name='Replaces existing taxonomy strings with those provided',
+    description=('This method allows the user to edit taxonomy strings. '
```
nbokulich, line 500: I recommend trimming back the description — it is a bit too specific/applied to a specific example. I recommend using a tutorial/usage example to give this example, but it is TMI for the description

rescript/plugin_setup.py
```diff
@@ -464,6 +466,50 @@
...
+    parameters={
+        'taxonomy_replacement_map': MetadataColumn[Categorical],
+        'num_expected_ranks': Int % Range(1, None),
+    },
+    outputs=[('edited_taxonomy', FeatureData[Taxonomy])],
+    input_descriptions={
+        'taxonomy': 'Taxonomy strings data to be edited.'
+    },
+    parameter_descriptions={
+        'taxonomy_replacement_map': 'A tab-delimitad metadata file in which '
```
nbokulich, line 483: maybe just call this "replacement_map" or something shorter to type?

Also I recommend dropping the example — as noted below, a tutorial would be a better place to communicate that specific example. (That example is also a bit controversial)

rescript/replace_taxonomy.py
```diff
@@ -0,0 +1,50 @@
...
+#
+# The full license is in the file LICENSE, distributed with this software.
+# ----------------------------------------------------------------------------
+
+from qiime2.plugin import MetadataColumn, Categorical
+import pandas as pd
+import re
+
+
+def make_substitutions_dict(taxonomy_replacement_map):
```
nbokulich, line 14: IMO it looks like this function can probably be removed... the `to_series()` method seems unnecessary now that you use a metadata column as input (it will already be a series), so this function just transforms the series to a dict, which could probably happen in-line below.

rescript/replace_taxonomy.py
```diff
@@ -0,0 +1,50 @@
...
+    replacement_dict = taxonomy_replacement_map.to_series().to_dict()
+    return replacement_dict
+
+
+def make_regex(substitutions):
+    strings = substitutions.keys()
```
nbokulich, line 20: if you only need the keys, then there is no need to transform to a dict above, just do:
```
strings = taxonomy_replacement_map.index
```
but there might be a more efficient way to do this (see below)

rescript/replace_taxonomy.py
```diff
@@ -0,0 +1,50 @@
...
+
+def replace_taxonomy(taxonomy: pd.Series,
+                     taxonomy_replacement_map: MetadataColumn[Categorical],
+                     num_expected_ranks: int = 7,
+                     ) -> pd.Series:
+
+    rd = make_substitutions_dict(taxonomy_replacement_map)
+    mr = make_regex(rd)
+
+    updated_tax_series = taxonomy.apply(lambda taxonomy:
```
nbokulich, line 40: why use re.sub for this at all? Wouldn't [pd.Series.replace](https://pandas.pydata.org/docs/reference/api/pandas.Series.replace.html) work? It uses `re.sub` under the hood so might save some effort (and potentially make the code more efficient), and looks like it can take a regex, dict, series, or other input to specify the substitutions. Then the preceding lines/functions could be dropped entirely, and just do something like:
```
updated_tax_series = taxonomy.replace(taxonomy_replacement_map)
```

rescript/replace_taxonomy.py
```diff
@@ -0,0 +1,50 @@
...
+                                        mr.sub(lambda match:
+                                               rd[match.group(0)], taxonomy))
+
+    irl = lineages_with_invalid_number_of_ranks(updated_tax_series,
```
nbokulich, line 44: is this just a safety catch to make sure that e.g., a greedy regex does not blow up the taxonomy?

there might be cases where this does not work, e.g., in many cases this function might be applied to taxonomic classifications of unknown sequences, which might have uneven numbers of ranks. So at the very least we should allow users to disable this function and place it in an `if` statement (`if num_expected_ranks:...`)

Do you have a good example of when this option is needed? if not, maybe it is unnecessary at this stage.

Another way to approach this more dynamically could be to count the number of ranks before and after replacement, instead of specifying a fixed length, like so:
```
ranks_in = taxonomy.str.count(';')
ranks_out = updated_tax_series.str.count(';')
if not ranks_in.compare(ranks_out).empty:
    raise ValueError("blah blah blah")
```

but again unless if you have a specific use case I wonder if this check is even needed (or could be moved to a separate validation function, e.g., to check that a taxonomy has even ranks)

rescript/replace_taxonomy.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def lineages_with_invalid_number_of_ranks(updated_tax_series,
+                                          num_expected_ranks):
+    invalid_lineages = [label for label, lineage in updated_tax_series.items()
+                        if len(lineage.split(';')) != num_expected_ranks]
+    return invalid_lineages
+
+
+def replace_taxonomy(taxonomy: pd.Series,
```
nbokulich, line 32: add an option to input a regex string? this might take more work so could also save for another day...

rescript/plugin_setup.py
mikerobeson: Agreed.

rescript/replace_taxonomy.py
mikerobeson: 'cause I'm a dummy 😄  . I just had to add `regex=True` n order for it to work on substrings. Otherwise it'll try and match the whole string. I got this working, and added a flag to toggle `regex`.

I removed all the other code too.

rescript/replace_taxonomy.py
mikerobeson: I just imagined many users might produce invalid taxonomy files. Which they won't be aware of until they proceed to downstream steps.

I came across this myself, when the best way to find a pattern was to include the `;` in the search string... and then I forgot to put it back with the replace string. Then I was missing a rank-level for that lineage.

I like your dynamic idea, particularly with the classification output! I'll work on implementing this route.

rescript/replace_taxonomy.py
nbokulich: oh good find... maybe this option should be exposed so that users can define whether to match entire or partial strings.

Is there no way to to substring substitution while `regex=False`? I worry a little bit about users unintentionally using wildcards that should be interpreted literally, e.g., brackets or punctuation in some taxonomies...

maybe your approach is more sound after all if it gives more flexibility...

rescript/replace_taxonomy.py
mikerobeson: Adding and setting as default.

rescript/replace_taxonomy.py
nbokulich: oooh very good example... I agree, some output validation is in order (to check for empty strings if not equal lengths)... I think the "dynamic" approach should scratch that itch...

one remaining question is whether we want to allow users to bypass this check, e.g., if they intentionally remove a rank for some reason... I am not sure when/why someone would want to, but something to chew on.

rescript/replace_taxonomy.py
mikerobeson: t appears I need to use `to_series()` and `to_dict()`, as outlined in the [MetadataColumn docs](https://dev.qiime2.org/latest/api-reference/metadata/#qiime2.metadata.MetadataColumn) for this to work. I assume I am missing something subtle here?

rescript/replace_taxonomy.py
mikerobeson: >Is there no way to to substring substitution while regex=False? 

I think we'd have to use `str.replace` or something along those lines. Then use a `lambda` function...

rescript/replace_taxonomy.py
mikerobeson: I have implemented a `warnings.warn`... But that'll only come up with the user runs `--verbose`. I can probably to a `try-except-finally`, so that it'll appear regardless of the `--verbose` flag being used?

rescript/replace_taxonomy.py
mikerobeson: >I worry a little bit about users unintentionally using wildcards that should be interpreted literally, e.g., brackets or punctuation in some taxonomies...

Yeah, me too, which is why I was using `re.escape`. But not sure how to implement that with the `pd.Series.replace`. I was initially thinking of having a flag that would toggle `re.escape` and `re`.

rescript/replace_taxonomy.py
nbokulich: yeah you are right, the metadata column does need to be converted to series. to_dict() might not be needed depending on whether you use pd.Series.str.replace or re.sub as discussed below.

rescript/replace_taxonomy.py
mikerobeson: I figured out a way to pass find-replace patterns to `pd.replace()` in the way I was imagining. We can now toggle whether or not the find pattern passes through `re.escape` prior to being sent to `re.compile`.  That resulting dictionary of compiled regexes can then be fed to `pd.replace()`. 🎉

rescript/plugin_setup.py
```diff
@@ -483,20 +485,44 @@
         'replacement_map': 'A tab-delimitad metadata file in which '
                            'the rank-lineage string in the \'id\' '
```
nbokulich, line 486: maybe just say "string" here and below? After all, in theory this could be used to modify all sorts of information (e.g., rank labels), not just rank lineage information.

rescript/plugin_setup.py
```diff
@@ -483,20 +485,44 @@
...
                            'column is replaced by the rank-lineage '
```
nbokulich, line 487: "column is replaced by the rank-lineage string. All string in a second named column."

Maybe change to:
"column is replaced by the replacement string(s) listed in the second column."

rescript/plugin_setup.py
```diff
@@ -483,20 +485,44 @@
...
+                 'This is often used to fix inconsistent and/or '
+                 'inccorect taxonomic annotations. The user can either '
+                 'provide two separate lists of strings, i.e. '
+                 '\'search-strings\', and \'replacement-strings\', '
+                 'on the command line, or a single tab-delimited '
+                 'replacement map file containing a list of these strings. '
+                 'In both cases the number of search strings must match the '
+                 'number of replacement strings. That is the first string '
+                 'in \'search-strings\' is replaced with the first string '
+                 'in \'replacement-strings\', and so on. Finally,if both '
```
nbokulich, line 523: does one or the other need to take precedence? maybe you could just merge the replacement dictionaries to allow use of both?

rescript/replace_taxonomy.py
```diff
@@ -9,15 +9,52 @@
 from qiime2.plugin import MetadataColumn, Categorical
 import warnings
 import pandas as pd
+import re
+from qiime2.plugin import List
+
+
+def make_search_replace_dict(search_strings, replacement_strings):
```
nbokulich, line 16: nice! 😎

rescript/replace_taxonomy.py
```diff
@@ -9,15 +9,52 @@
...
+    # a literal string match.
+    new_rm_dict = {(re.compile(k) if use_regex else re.compile(
+                            re.escape(k))): v for k, v in rm_dict.items()}
+
+    # Note, becuase we are compiling a regex either with or without
+    # `re.escape`, we must always set `regex=True` for `pd.replace`.
+    updated_tax_series = taxonomy.replace(new_rm_dict, regex=True)
 
     # User sanity check. Though only appears with `--verbose`
     ranks_in = taxonomy.str.count(';')
```
nbokulich, line 60: `ranks_in.compare(ranks_out)` will return a Series of the diff between `ranks_in` and `ranks_out`, with an index corresponding to the row number... so instead of merely warning that the number of ranks differ, we could actually specify which lines differ.

I am not saying that this needs to be done or would even be useful... just throwing the idea out there in case you think it is useful.

rescript/tests/data/taxonomy-replacement-regex.txt
```diff
@@ -0,0 +1,6 @@
+id	replacements
+p__.*; c__	p__LAME-PYHLA; c__
+s__uncultured	s__UNCIVILIZED
```
nbokulich, line 3: lol!

nbokulich (changes requested):
thanks @mikerobeson ! this is looking good. I have a few high-level comments in-line.

Also, could I request a few other unit tests? You have already done heaps, but I see a couple other possible tests:
1. test that the warning is raised when the number of ranks in/out do not match
2. some more edge cases should/could be tested, to see what happens if someone "accidentally" mangles the taxonomy with a wildcard in their regex. this might also be combined with point 1 (i.e., mangling and loss of ranks might be one and the same).

Re: point 2 above, it is not clear to me that any validation is done to check for empty taxonomies in the output. Should this be done, to make sure that mangling leading to empty taxonomy does not occur? Empty taxonomies can cause errors downstream, so a quick validation might be in order. (though a very strong argument could be made that this should be added to the format validation to disallow empty taxonomies in the first place)

Thanks!

rescript/replace_taxonomy.py
mikerobeson: I like this and implemented it into the warning. Trying to figure out how to test that the warning is raised.

rescript/plugin_setup.py
mikerobeson: I think we'll need to workout why passing in regex strings at the command line require `\\` escape characters, while the regex patterns in the mapping file use `\` to escape, before we merge these.

mikerobeson: It will now merge the dictionaries if the user supplies the stings via the mapping file and the command line.

rescript/replace_taxonomy.py
nbokulich: you can test the the warning is raised with [assertWarnsRegex](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertWarnsRegex)

here's an example from [q2-sample-classifier](https://github.com/qiime2/q2-sample-classifier/blob/a787ed020e7fc5e4bd0b045b563025d6bec9cf59/q2_sample_classifier/tests/test_estimators.py#L257)

nbokulich (changes requested):
thanks @mikerobeson ! I was not sure if this was ready for review again or still in progress but I saw a couple little things float up that I thought I would comment on before you carried them further... basically just streamlining the code (tests mostly) to reduce redundancy, see in-line comments. Thanks!

rescript/edit_taxonomy.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def edit_taxonomy(taxonomy: pd.Series,
+                  replacement_map: MetadataColumn[Categorical] = None,
+                  search_strings: List = None,
+                  replacement_strings: List = None,
+                  use_regex: bool = False,
+                  ) -> pd.Series:
+
+    if search_strings and replacement_strings and replacement_map:
```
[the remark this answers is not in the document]
mikerobeson: No worries. I'll take it! Thank you for doing this! :-)

rescript/tests/test_edit_taxonomy.py
```diff
@@ -0,0 +1,332 @@
...
+                               'g__Escherichia-Shigella; s__'),
+                    'Shig01': ('d__Bacteria; p__LAME-PYHLA; '
+                               'c__Gammaproteobacteria; o__Enterobacterales; '
+                               'f__Enterobacteriaceae; '
+                               'g__Escherichia-Shigella; s__')}
+
+        self.maxDiff = None
+        self.assertEqual(obs_dict, exp_dict)
+
+    def test_edit_taxonomy_pass_regex_map_and_cl(self):
```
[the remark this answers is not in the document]
mikerobeson: Oh right! Not sure why I did not realize that myself. Sorry! 🤦

nbokulich (approved):
looks great, thanks @mikerobeson !

Issue #119, opened 2021-10-01
Request for function to reverse complement


jwdebelius:
I would really like the ability to reverse complement sequences in QIIME 2 without re-orienting against a reference. Is this appropriate for RESCRIPt (if so, I'm happy to issue the PR), or is it better for another more central plugin?

nbokulich:
I'd be in favor of adding this as an option to `orient-seqs`. E.g., make the reference optional and expose an `orientation` parameter to reverse, complement, or reverse-complement? (maybe that's overthinking it/way too many options)

Or just a separate method is fine too... what do you think @mikerobeson ?

mikerobeson:
I think would be a great idea @jwdebelius! I think I'd recommend it as an addition to `orient-seqs`, with the options that @nbokulich recommended, *i.e.* reverse, complement, or reverse-complement. I do not think this would create a problem of "too many parameters", I mean look at the iqtree plugin. 😄 

🤔 Would it be worth also implementing the ability to specify feature IDs to be reverse-complimented. That is, there have been times when I only needed to reverse-compliment a handful of sequences within my dataset. But I guess this'd argue for a separate action? If so, perhaps we can rename the `orient-seqs` action as `orient-seqs-against-reference` and `reverse-compliment`? Maybe I am making things too complicated now. 

But I am for adding to `orient-seqs`.

nbokulich:
hey @jwdebelius any follow-up thoughts on this? have you already implemented this elsewhere? this would be a great addition to `orient-seqs` if you are still interested!

@mikerobeson the option to re-orient specific feature IDs is a neat idea! I do worry that could over-complicate the method though. A more transparent (though less convenient) work-around might be to filter to keep those IDs, re-orient with this method, then merge back in. Maybe let's start with the "simple" implementation and we could add this option later on if it is a common enough need.

jwdebelius:
Sorry, I've been weirdly scattered. Yes, I have a version implemented in sidle right now, but could easily move it over. My initial pass is pretty simple. I think I can move it into orient-seqs, though!

PR #121, opened 2021-11-10
Adds reverse complementing to orient-seqs

jwdebelius:
As discussed in #119. 

Probably worth mentioning that I see some utility in keeping reverse complemented for aligned sequences, which is whats currently implemented, but I haven't explored the implications for orienting to a reference where it probably _does_ break vserach, and might be a motivation for a separate method.


nbokulich:
Hi @jwdebelius thanks for this! 🎉 

Some initial ideas re: the point you raise about aligned seqs

I agree with you: reverse orientation of aligned sequences is a prominent use case, and we should support. But we do not want to introduce a bug re: VSEARCH compatability.

This might be a case for using `TypeMap` — i.e., so that `FeatureData[AlignedSequence]` is only accepted as input when `reference_sequences=None`. If that works, we will need to make the usage clear in the help documentation.

We could also introduce a separate `orientation` parameter (with options same, complement, reverse, or reverse complement), which might be a bit easier to use with `TypeMap`, though only if we think these options would be useful. 

Another idea that might be cleaner and more functional. Accept aligned seqs as input with an unaligned ref.
1. degap a copy of the aligned seqs on input
2. run the degapped seqs through vsearch. Vsearch reports the orientation of each seq (+ / -)
3. orient the original aligned seqs based on the sign, and output

No matter which route we take, could you please add unit test(s) for orienting aligned seqs with the various options? **I think that the current setup actually will not work with aligned sequences, since the view type is `DNAFASTAFormat`... I believe a `DNAIterator` is needed instead.**

@mikerobeson any thoughts?

mikerobeson:
For whatever reason I just noticed this. I'll look into it now. 

I do not currently have much to @nbokulich's comments. Although @jwdebelius and I were discussing the old [filter_alignment.py](http://qiime.org/scripts/filter_alignment.html) script from QIIME 1. In particular, the ability to remove poorly aligned sequences using the `--remove_outliers` and associated flags. Perhaps we can add a separate action in either the `q2-alignment` plugin, or rescript simply called `detect-alignment-outliers`?

This would set us up for several downstream use cases: 
1) detect the poorly aligned sequences (write the sequences and/or their ids to file), then investigate / filter downstream, using one of the QIIME 2 standard filtering commands
2) use output from 1, to specify which sequences in the FASTA file should be reverse-complemented, leave the others as is.

nbokulich:
that sounds very useful! also a very different use case, i.e., for filtering alignments on quality (which sounds like a method for q2-alignment). Sounds like the q1 method also masked the alignments? So a very different beast (and one that is partially done now in q2-alignment?). Either way, filtering alignments sounds like very different functionality from read orientation.

So I propose that
1. we stick with the original plan to support [aligned] seq orientation in rescript
2. you open a separate issue for alignment filtering in q2-alignment @mikerobeson

@jwdebelius how would you like to proceed? See my note above about your current PR most likely not working for aligned seqs... I propose two options:
1. you add fully functional support for alignment orientation, incl. unit test(s) (see my ideas above)
2. you drop alignment support from the current PR. We can address this in a future PR.

I am happy with either path 😁

jwdebelius:
Hi @nbokulich,

Sorry, things have been umm... busy the last 3 months. I'm a fan of the second option. I'll clean up the PR, and hopefully you can check it and it can be merged.

nbokulich:
hi @jwdebelius, thanks for updating this PR! Please let me know when this is ready to review

jwdebelius:
Thanks @nbokulich, I think it should be good to go!

LICENSE
```diff
@@ -1,6 +1,6 @@
 BSD 3-Clause License
 
-Copyright (c) 2021, QIIME 2 development team.
+Copyright (c) 2022-2022, QIIME 2 development team.
```
nbokulich, line 3: 2022-2022 looks weird...

why not 2021-2022 or just 2022? (I realize this is not your problem, it's issues with q2lint and us not yet recognizing the new year, so we can fix if this is a hassle)

rescript/orient.py
```diff
@@ -1,55 +1,68 @@
...
-               '--db', str(reference_sequences), '--id', str(perc_identity),
-               '--maxaccepts', '1', '--strand', 'both', '--qmask', 'none',
-               '--query_cov', str(query_cov), '--threads', str(threads),
-               '--userfields', 'qstrand', '--userout', out.name]
-        if left_justify:
-            cmd.append('--leftjust')
-        run_command(cmd)
-        with open(out.name, 'r') as orient:
-            orientations = [line.strip() for line in orient]
+    if (reference_sequences is not None):
```
nbokulich, line 23: the parenthesis should not be needed. Or am I missing something?

rescript/orient.py
```diff
@@ -1,55 +1,68 @@
...
+                cmd.append('--leftjust')
+            run_command(cmd)
+            with open(out.name, 'r') as orient:
+                orientations = [line.strip() for line in orient]
 
-    # if any query seqs are in reverse orientation, reverse complement
-    if '-' in orientations:
+        # if any query seqs are in reverse orientation, reverse complement
+        if '-' in orientations:
+            matched = DNAFASTAFormat()
```
nbokulich, line 52: move these two lines to the top of the function outside of the conditional statement? then you avoid duplicating the lines below.

rescript/orient.py
```diff
@@ -1,55 +1,68 @@
...
+        # report orientation of query seqs relative to reference seqs.
+        with tempfile.NamedTemporaryFile() as out:
+            # note: qmask is disabled as DNAFASTAFormat requires all output
+            # seqs to be uppercase. Could loop through output seqs to convert
+            # to upper but which is faster: disabling masking or looping
+            # through with skbio?
+            cmd = ['vsearch',
+                   '--usearch_global', str(sequences),
+                   '--matched', str(matched_temp),
+                   '--notmatched', str(notmatched),
```
nbokulich, line 34: what do you think of raising a warning message or Note that these various parameters are ignored when running in reverse complement mode? (this info can also just be described in the parameter/action descriptions)

rescript/plugin_setup.py
```diff
@@ -550,13 +550,16 @@
              ('unmatched_seqs', FeatureData[Sequence])],
     input_descriptions={
         'sequences': 'Sequences to be oriented.',
-        'reference_sequences': 'Reference sequences to orient against.'},
+        'reference_sequences': ('Reference sequences to orient against. If no'
+                                ' reference is provided, all the sequences '
```
nbokulich, line 554: can the space at the beginning of this line be moved to the end of the previous line? (makes for easier reading)

nbokulich (changes requested):
thanks @jwdebelius ! This looks great — I just have a few minor comments inline. Thanks!

LICENSE
jwdebelius: I hope this works with the linting

rescript/orient.py
jwdebelius: That doesn't work. We only need to apply this if we're using reverse complementing against a reference. If thats the case, we check for a reverse orientation (line 51), then loop through the sequences (54-59) and if a given sequence is negative, reverse complement that. The if statements are on _slightly_ different variables. I could re-name the variables if that would help?

I think this is an indented version of your original logic, just brought under the "has reference" umbrella

rescript/orient.py
jwdebelius: I tried that in Sidle, and the warning messages weren't raised unless you were in verbose mode. See this forum conversation with Matt:

https://forum.qiime2.org/t/warning-not-raised/21712/

The challenge with adding it to the docs is that right now, RESCRIPt sets a global variable for vsearch parameters that's used for parameters here. I tried to add an update to the docs, let me know if you that thats sufficient or if an expanded doc version needs to be built that adds a warning string to the parameters.

jwdebelius:
I hope this addresses your concerns @nbokulich! There are a couple of things I wasn't sure how to address directly

rescript/plugin_setup.py
```diff
@@ -550,9 +550,11 @@
              ('unmatched_seqs', FeatureData[Sequence])],
     input_descriptions={
         'sequences': 'Sequences to be oriented.',
-        'reference_sequences': ('Reference sequences to orient against. If no'
-                                ' reference is provided, all the sequences '
-                                'will be reverse complemented.')},
+        'reference_sequences': ('Reference sequences to orient against. If '
+                                'no reference is provided, all the sequences '
+                                'will be reverse complemented and all '
+                                'parameterswill be ignored.'
```
nbokulich, line 556: missing space in "parameterswill" (or is that a kind of beverage? 😁 )

rescript/orient.py
nbokulich: you're right. Technically ```matched = DNAFASTAFormat()``` could be moved outside of the logic, and it is repeated on line 63... but the way you have it is more readable anyway (better than saving 1 line of code 😆 )

rescript/orient.py
nbokulich: Good points, and the changes you already made to the input descriptions do most of what's needed.

Want to just add an extra 1-2 sentences to the action description in `plugin_setup.py`? Something like this (added to the end of the description):

> Alternatively, if no reference sequences are provided as input, all input sequences will be reverse-complemented. In this case, no alignment is performed, and all alignment parameters (`perc_identity`, `query_cov`, `threads`, and `left_justify`) are ignored.

nbokulich (changes requested):
thanks @jwdebelius !

Almost done... pls see my response and comment in-line.

rescript/plugin_setup.py
jwdebelius: 🍻

jwdebelius:
I think this should address your documentation issues, @nbokulich.

nbokulich (approved):
looks great, thanks so much for this PR @jwdebelius !

PR #130, opened 2022-03-23
drop experimental loo cv method

nbokulich:
there is an experimental loo cv method that was never exposed as an action.

this pulls some functions from q2-feature-classifier that are being cut out in this PR: https://github.com/qiime2/q2-feature-classifier/pull/176

so I propose just dropping this function in rescript for now. We can always add back later if anyone wants... does that sound reasonable @mikerobeson ?

### DIFF
--- a/rescript/cross_validate.py
+++ b/rescript/cross_validate.py
@@ -11,8 +11,6 @@ import qiime2 as q2
 import timeit
 from warnings import filterwarnings
 from sklearn.model_selection import StratifiedKFold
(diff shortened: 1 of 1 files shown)


mikerobeson:
Sounds good to me @nbokulich. I can merge if there is nothing else to modify.

nbokulich:
thanks! nothing else needed IMO, ready to merge. I can open an issue to "bookmark" this for a possible later date.

PR #139, opened 2022-06-22
dep/bug: missing vsearch dependency

thermokarst:


thermokarst:
Looks like the mac build picked up a newer version of vsearch, while the linux build picked up an older one. I'll work on making the necessary tweaks here.

Issue #136, opened 2022-06-07
ENH: add vsearch's `-qsegout` option to RESCRIPt


mikerobeson:
We've been discussing internally, for quite a while now, to be able to extract target / amplicon regions from online sequence repositories in which one or more of the PCR primers are not present or have been removed prior to that reference sequence being deposited. That is, using primers to search for and extract the target region will fail. Or better yet, to be able to extract target regions from complete genomes. Anyway, I just realized I never made an issue for this, so here it is. :-)

This, in the past, was eloquently solved by the [-qsegout](http://drive5.com/usearch/manual8.1/opt_qsegout.html) (and perhaps [-tsegout](https://drive5.com/usearch/manual8.1/opt_tsegout.html)?) command of usearch. That is, it can use actual amplicon data, or a pre-generated list of sequences, to use for searching and extracting the target amplicons from a larger reference database (i.e. GenBank, BOLD, etc...). [This](https://www.drive5.com/usearch/manual8.1/utax_user_train.html) is the exact approach I used to make reference databases for my [pig diet paper](http://dx.doi.org/10.1002/ece3.3638), and have been wanting to implement this approach ever since we started development of RESCRIPt. Now that this has been [implemented in vsearch](https://github.com/torognes/vsearch/issues/447). I think we can start planning to include this functionality. 

*Although I think we may be limited to implementing the vsearch version which now includes the `-qsegout` option, as vsearch is currently pinned within QIIME 2 due to another plugin dependency? Hopefully, that'll be resolved soon.*

mikerobeson:
Just a short little update... I installed the latest version of vsearch (2.21.1) into my QIIME 2 test environment. I was able to run `core-diversity-metrics` and `deblur denoise-16S` w/o issue. Thus, on the surface, it appears that UniFrac and deblur may no longer be the culprits? I can try running the other deblur actions, etc... Are there are plugins / actions I should look into? Let me know.

thermokarst:
Hey @mikerobeson - I don't think deblur specifically _pinned_ vsearch, but rather just didn't work with newer version of vsearch. A new release of deblur just landed, we're working on pulling it into the core distro, once that's set I hope we'll be able to unpin vsearch.

mikerobeson:
Thanks for the update @thermokarst! 

[...]

PR #140, opened 2022-06-22
ENH: add vsearch's --qsegout option to extract reads

mikerobeson:
In my initial #138 PR, I had mistakingly branched from #133 somehow. 🤦 So, I reset and tried again. 

Anyway, this addresses #136 

-Mike


mikerobeson:
Yeah, I'm looking at that now ... 🤔

mikerobeson:
Yeah, not sure why the Linux version keeps installing vsearch 2.7 instead of 2.21.

thermokarst:
@mikerobeson - you should coordinate that version pin with @ebolyen - he might want to add that to our shared conda build config.

mikerobeson:
Thanks @thermokarst! I was just testing if I can force the pin here for a sanity check. But yes, @ebolyen if we make sure that the latest version of vsearch (2.21.1) is pulled in.

misialq (approved):
Hey @mikerobeson, thanks for adding that! The code looks good + all runs as expected (I played around with the data you provided for tests by changing the sequences a bit and adjusting the identity thresholds - everything ran nicely) 🚀 🥇 

A random thought: what should be the expected behaviour if one of my query sequences   contains more than one of the references? (not sure how likely that is though, so please ignore the question if it doesn't apply here). In my test it seemed to extract the sequence that happened to be the first in the query (which I personally sort-of expected).

mikerobeson:
>A random thought: what should be the expected behaviour if one of my query sequences contains more than one of the references?... In my test it seemed to extract the sequence that happened to be the first in the query (which I personally sort-of expected).

Great question! Yes. It'll return the first instance it encounters. We could potentially pull more by setting `--strand 'both'`, but the problem is that multiple sequences will be written out with the same exact sequence ID, which will then cause the code to fail as `FeatureData[Sequence]` expects unique sequence IDs. This is why I am using the `--strand 'plus'`.  Probably something we can improve upon in a later update?

I figure with the plethora of amplicon and genomic sequences available for most common amplicon targets, this might be not be a major issue as we'll capture most things. 🤷 

If you're okay with this, go ahead and merge. :-)

misialq:
Great, thanks for the explanation @mikerobeson! I agree that this does not seem like a major issue and, as you said, we can always follow up later, if necessary. Will merge! ✅

PR #142, opened 2022-08-29
update readme for q2-2022.8

mikerobeson:
This PR updates the README:
 - Minimum version of QIIME 2 is now `2022.8`
 - Added link to the new [Using RESCRIPt's 'extract-seq-segments' to extract reference sequences without PCR primer pairs.](https://forum.qiime2.org/t/using-rescripts-extract-seq-segments-to-extract-reference-sequences-without-pcr-primer-pairs/23618) tutorial.
 - Added link to the related [db-benchmarks-2020 repo](https://github.com/bokulich-lab/db-benchmarks-2020).


mikerobeson:
I'm not sure why the build is failing now. Any thoughts?

PR #149, opened 2023-01-10
MAINT: update `orient-seqs` to use vsearch `orient`

### LINKED ISSUE #146: MAINT: update `orient-seqs` to use vsearch `orient`
`orient-seqs` currently uses `vsearch --usearch_global` to search a database, then reorient.

This was always a short-term solution, as vsearch was previously pinned across many Q2 plugins so we could not use the more recent `vsearch --orient` option. But that pin is now gone.

We should update to use `orient`. Advantages:
1. cut down on current code in `orient-seqs`
2. probably more efficient
3. can operate on both FASTA and FASTQ data

The disadvantage is that some of the options available in `--usearch_global` might not be available in `orient`.

### LINKED ISSUE #151: REVERT future warning for orient-seqs
REVERT this: https://github.com/bokulich-lab/RESCRIPt/pull/150

only after this PR is merged and the deprecation period is complete: https://github.com/bokulich-lab/RESCRIPt/pull/149

ETA: May 2023

mirand863:
This PR replaces `vsearch --usearch_global` with `vsearch --orient`, which is more efficient (closes #146). Additionally, it removes the deprecation warning inserted in version `2023.2` (closes #151).

Some parameters were removed since they are not supported with `--orient`:

- `--id`
- `--maxaccepts`
- `--strand`
- `--query_cov`
- `--userfields`
- `--leftjust`

However, optional parametes available with `--orient` were added:

- `--dbmask`
- `--relabel`
- `--relabel_keep`
- `--relabel_md5`
- `--relabel_self`
- `--sizein`
- `--sizeout`

### DIFF
--- a/rescript/orient.py
+++ b/rescript/orient.py
@@ -25,6 +25,8 @@ def orient_seqs(sequences: DNAFASTAFormat,
         # seqs to be uppercase. Could loop through output seqs to convert
         # to upper but which is faster: disabling masking or looping
         # through with skbio?
+        # NOTE: --id, --query_cov and --leftjust are disabled since
+        # they are not available with orient
         cmd = [
             'vsearch',
             '--orient', str(sequences),
--- a/rescript/orient.py
+++ b/rescript/orient.py
@@ -21,8 +21,7 @@ def orient_seqs(sequences: DNAFASTAFormat,
                 ) -> (DNAFASTAFormat, DNAFASTAFormat):
     matched, notmatched = DNAFASTAFormat(), DNAFASTAFormat()
     if reference_sequences is not None:
-        # use vsearch to search query seqs against reference database
-        # report orientation of query seqs relative to reference seqs.
+        # use vsearch to orient seqs against reference database
         with tempfile.NamedTemporaryFile() as out:
             # note: qmask is disabled as DNAFASTAFormat requires all output
             # seqs to be uppercase. Could loop through output seqs to convert
--- a/rescript/orient.py
+++ b/rescript/orient.py
@@ -19,7 +19,7 @@ def orient_seqs(sequences: DNAFASTAFormat,
                 threads: int = 1,
                 left_justify: bool = False,
                 ) -> (DNAFASTAFormat, DNAFASTAFormat):
-    matched_temp, notmatched = DNAFASTAFormat(), DNAFASTAFormat()
+    matched, notmatched = DNAFASTAFormat(), DNAFASTAFormat()
     if reference_sequences is not None:
         # use vsearch to search query seqs against reference database
         # report orientation of query seqs relative to reference seqs.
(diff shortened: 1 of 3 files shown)


mirand863:
@nbokulich,
I opened an issue on vsearch and the reason that these parameters are missing is because usearch also does not have them. However, I believe some parameters are irrelevant when using orient (`--maxaccepts`, `--strand` and `userfields`). The other parameters can be computed if an alignment is provided. Do you think it would be worth adding alignment information or would it be fine to remove `--id`, `--query_cov` and `--leftjust`? I also looked into the tabbed output provided by orient, but it does not have enough information to be able to replace these parameters.

rescript/tests/test_orient.py
```diff
@@ -66,6 +66,21 @@ def test_reorient_default(self):
         exp_unmatched_ids = {'JUNK', 'MOREJUNK'}
         self.assertEqual(unmatched_ids, exp_unmatched_ids)
 
+    def test_reorient_no_ref(self):
```
nbokulich, line 69: why was this test moved? It does not look like anything changed.

nbokulich (changes requested):
thanks @mirand863 ! Looks good to me, I have only one comment in-line. @misialq will also review/test locally. 

I think that we should drop these parameters if they are not used by `vsearch orient`.  I agree that they are probably not needed by users (although there could be some use for `id` and `query_cov` in theory). @mikerobeson do you agree with this?

However, if we make breaking changes we should go through a short deprecation cycle, so I am tagging this as `do-not-merge` for now. Once approved, we can merge in a future release of rescript.

rescript/tests/test_orient.py
mirand863: I moved it to keep the commented tests in the end, but it was unnecessary. Now I just added decorators to skip the failing tests, so in the future if anything changes they can be executed again by simply removing the decorators.

mirand863:
> thanks @mirand863 ! Looks good to me, I have only one comment in-line. @misialq will also review/test locally.
> 
> I think that we should drop these parameters if they are not used by `vsearch orient`. I agree that they are probably not needed by users (although there could be some use for `id` and `query_cov` in theory). @mikerobeson do you agree with this?
> 
> However, if we make breaking changes we should go through a short deprecation cycle, so I am tagging this as `do-not-merge` for now. Once approved, we can merge in a future release of rescript.

A deprecation cycle is a great idea.

I would also like to suggest using [semver](https://semver.org/) for versioning instead of the release date. With semver it becomes easier to warn users that a new version is incompatible with the previous one by marking as a major update. Moreover, tagging a new version can be automated with [github actions](https://github.com/anothrNick/github-tag-action), where the bumping can be done according to pull requests' titles, e.g., by adding the tags #major or #minor to the title.

rescript/tests/test_orient.py
mikerobeson: I agree with @nbokulich, any parameters that are not available for `vsearch orient` should be removed, along with adding any potentially useful parameters for the `orient` command outlined in the [manual](https://github.com/torognes/vsearch/releases/download/v2.22.1/vsearch_manual.pdf).

rescript/orient.py
```diff
@@ -19,50 +18,34 @@ def orient_seqs(sequences: DNAFASTAFormat,
                 threads: int = 1,
                 left_justify: bool = False,
                 ) -> (DNAFASTAFormat, DNAFASTAFormat):
-    matched_temp, notmatched = DNAFASTAFormat(), DNAFASTAFormat()
+    oriented, notmatched = DNAFASTAFormat(), DNAFASTAFormat()
```
mikerobeson, line 21: Just wanted to point out that `vsearch orient` can handle both fastq and fasta files. This can likely wait for a separate PR. But we'd have to figure out a way to handle paired-end reads. That is, if we orient the forward read, then the reverse read should be oriented too.

mirand863:
@nbokulich,

Added a new [PR #150](https://github.com/bokulich-lab/RESCRIPt/pull/150) to warn users of a future deprecation of the parameters discussed here.

misialq (commented):
Hey @mirand863, I gave it a go and all works as expected, thanks! I have two more cosmetic comments though:
- Could you adjust the action description in the `plugin_setup.py`?  I think the last section becomes irrelevant after dropping the parameters.
- I would suggest just removing the tests that are not being used. They are addressing functionality that was removed and most likely will not be brought back so no need to keep them.

rescript/tests/test_orient.py
mirand863: @mikerobeson Thank you for the suggestion! I just added the optional parameters available with `orient`.

mirand863:
> Hey @mirand863, I gave it a go and all works as expected, thanks! I have two more cosmetic comments though:
> 
>     * Could you adjust the action description in the `plugin_setup.py`?  I think the last section becomes irrelevant after dropping the parameters.
> 
>     * I would suggest just removing the tests that are not being used. They are addressing functionality that was removed and most likely will not be brought back so no need to keep them.

Thank you for the great suggestions! I removed the tests and edited the description in `plugin_setup.py` to remove old parameters and add the new ones.

Issue #47, opened 2020-06-12
add `get_gtdb_data` method


nbokulich:
the parallel of `get_silva_data` but download, import, and format from GTDB:
* FASTA sequences
* taxonomy
* reference phylogeny?

should add license information to main readme and method description, similar to what we've done with the SILVA license. 

GTDB is pretty much already Q2-compatible but this pipeline would streamline things for users, allow us to easily pipe these artifacts into classifier training workflows, and bake GTDB into provenance to ensure that users cite appropriately...

PR #154, opened 2023-02-15
ENH: initial draft of `get-gtdb`

mikerobeson:
This PR addresses #47, to download SSU data from [GTDB](https://gtdb.ecogenomic.org/).

The user provides the version of SSU data they wish to download from GTDB. Currently only versions `202` and `207` are allowed, with `207` being the default.

The code will download the sequence (.tar.gz) and taxonomy (.gz) files for the Archaea and Bacteria, then merge them together producing two output files of the types:
- FeatureData[Taxonomy]
- FeatureData[Sequence]


nbokulich:
awesome, thanks @mikerobeson ! Let me know when this is ready to review. Just a few early comments and questions pre-review, none of which are very important:
1. instead of automatically downloading and merging both bacteria and archaea, do you think that it would be useful to expose an option? E.g., for bacteria, archaea, or both? (both can be default)
2. I suppose right now it is just SSU. Do you think that an option (or separate action) to download the GTDB genomes (instead of SSU) would be useful? @misialq any thoughts?
3. GTDB also provides reference phylognies. Do you think that these would be a useful companion file to download? We discussed this previously re: `get_silva_data`. One issue I see is that GTDB only supplies separate trees for bacteria and archaea.

mikerobeson:
Hi @nbokulich, 

- RE 1: I was considering setting up an option for selecting `Archaea`, `Bacteria`, `Both`. But I was concerned how many users might simply download only bacteria, for classifying their bacterial data without any form of  'outgroup' taxa, *e.g.* the Archaea. I did not want to contribute to the issue we see often see with fungal ITS data. That is, when classifying fungi w/o non-fungi outgroup taxa, many non-fungal taxa can still be classified as fungi. I could offer this option though, if there are other use cases I'm not considering.
- RE 2: I can go either way with this. If needed, I can rename this action as `get-gtdb-ssu`, and then we can set up another `get-gtdb-genomes` action for the short-term. I've not looked into any of the existing data types for genome data, or what anyone else has done with genome data within QIIME 2.
- RE3: Yeah, the issue is the separate phylogenies, I am not sure how useful they'd be, or what they'd be used for...  Perhaps helping construct a GTDB based sepp reference?

nbokulich:
Hi @mikerobeson ,
1. yeah good point. I suppose I can't really think of a good use case, unless if the data are being used in some other way (e.g., for building a phylogeny but not for taxonomic classification). My gut is still telling me to give the user options, but my brain agrees that we don't want to make something that can be abused. Making `both` the default would limit accidental abuse.
2. Relevant types exist for genome sequence data in `q2-types-genomics`. @misialq do you think that this should be a separate action or exposed as an option (e.g., to grab SSU or genome seqs)? Either way, this can be a follow-up PR but we should decide now for optimal naming.
3. Yeah agreed, it's unfortunate. 😞

nbokulich (changes requested):
Hey @mikerobeson ,
Looking good! I have a few more comments inline, mostly on the tests.

I am also unable to get this to run at the moment (but am testing in a 2022.8 environment, so will try later in a fresh env). Have you confirmed that the different versions all download as expected on your end?

misialq:
Hey both,
just a quick comment on the SSU vs. genomes: I'm not sure what the genome data fetched from GTDB looks exactly like but I think it would probably make sense and would not be too complicated to just have it in the same action as most of the code would be shared between those. And yes, `q2-types-genomics` does have some `GenomeData` types, although those are more suited for storing genomic features (loci, genes, proteins) rather than full genomes (if it's just genomic nt sequences that are being fetched, aren't those anyway more like a `FeatureData[Sequence]` type?)

nbokulich:
>  (if it's just genomic nt sequences that are being fetched, aren't those anyway more like a FeatureData[Sequence] type?)

Yeah as far as I recall, these are just FASTAs of the whole genomes, so `FeatureData[Sequence]` should work.

rescript/get_gtdb.py
```diff
@@ -0,0 +1,138 @@
...
+                # taxonomy files are contained within ``.gz`
+                else:
+                    try:
+                        unzipped_destination = os.path.splitext(destination)[0]
+                        print('  Unzipping {0}...\n'.format(bn))
+                        _gzip_decompress(destination, unzipped_destination)
+                        destination = unzipped_destination
+                    except OSError:
+                        pass
+                    tax_results.append(qiime2.Artifact.import_data(
```
misialq, line 134: I think those two branches could become two little, independent functions which you could then easily write tests for.

rescript/get_gtdb.py
```diff
@@ -0,0 +1,138 @@
...
+    with tempfile.TemporaryDirectory() as tmpdirname:
+        for sttype, q_info in queries.items():
+            for domain, url, dtype, fmt in q_info:
+                print('Retrieving {0} for {1} from {2}'.format(
+                      sttype, domain, url))
+                # grab url
+                bn = os.path.basename(url)
+                destination = os.path.join(tmpdirname, bn)
+                try:
+                    urlretrieve(url, destination)
```
misialq, line 106: I agree with @nbokulich that no data should be fetched in the tests. If you refactor a little (see my other comment below) and then only mock this `urlretrieve` function, it should be quite straightforward to test this entire function :)

rescript/get_gtdb.py
```diff
@@ -0,0 +1,138 @@
...
+                    '%s.0/genomic_files_reps/%s_ssu_reps_r%s.tar.gz')
+    base_tax_url = ('https://data.gtdb.ecogenomic.org/releases/release%s/'
+                    '%s.0/%s_taxonomy_r%s.tsv.gz')
+
+    for version, dcp in ver_dom_dict.items():
+        for dom, cp in dcp.items():
+
+            queries['Taxonomy'].append(
+                (dom,
+                 base_tax_url % (version, version, cp, version),
```
misialq, line 72: Is there any reason to use this way of string formatting and the `format` method everywhere else below? Maybe you could use `format` here too, for consistency?

rescript/tests/test_get_gtdb.py
```diff
@@ -0,0 +1,58 @@
...
+                         'bac120_ssu_reps_r202.tar.gz')]
+        self.assertEqual(obs_tax_urls, exp_tax_urls)
+        self.assertEqual(obs_seq_urls, exp_seq_urls)
+
+        # test that these URLs work
+        for u in obs_tax_urls + obs_seq_urls:
+            try:
+                urlopen(u)
+            except HTTPError:
+                raise ValueError('Failed to open URL: ' + u)
```
misialq, line 58: This should also not be happening here. If this test fails in the CI (or anywhere else really), we still wouldn't know whether it was because e.g. their API was just unreachable temporarily or the URLs didn't exist or something yet different. I think it's enough that you assert that the URLs constructed by your function look like what you expect them to look like.

misialq (changes requested):
Hey @mikerobeson, I tried it out locally (all works!) and left a few more comments below.

rescript/get_gtdb.py
mikerobeson: Nice catch! No reason at all... just a momentary lapse... :-)

rescript/tests/test_get_gtdb.py
mikerobeson: I like this test, because some online repositories can change their file structure / urls. So, I want something to notify me that something is wrong. Even though this can fail for the other reasons you outlined. I can leave a more explicit warning, that the website might simply be down, and to try the test at a later time... We also use a similar test for the SILVA repository. But I am fine with removing this url check.

rescript/tests/test_get_gtdb.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def test_retrieve_data_from_gtdb(self):
```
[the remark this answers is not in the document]
mikerobeson: I've implemented the mock tests. This was my first time using this approach, so let me know if I've implemented this properl. 😄

nbokulich:
Hey @mikerobeson are your latest changes ready for review or are you still working on it? 

I did some user testing to check the outputs etc, and all works 👍 . However, I notice that there are many more records in the taxonomy (317543) than there are sequences (40660) (these counts are from the "Both" outputs). I confirmed that this matches the contents of the files from GTDB, so it looks like everything is operating as intended. But I wonder: do you know why there is this disparity?

rescript/tests/test_get_gtdb.py
```diff
@@ -0,0 +1,111 @@
...
+
+        def _makey_fakey_bact(faking_ignore_this):
+            return [self.bact_tax], [self.bact_seqs]
+
+        # default (both domains, version 207)
+        with patch('rescript.get_gtdb._retrieve_data_from_gtdb',
+                   new=_makey_fakey_both):
+            rescript.actions.get_gtdb_data(
+                version='207', domain='Both')
+            self.assertTrue(True)
```
misialq, line 99: I don't think this assertion is needed here 😅 Maybe instead you could just assert that the action produced two artifacts? (either by checking length, or even using something like `isinstance` to test that they are of the right type?) - same applies below.

rescript/tests/test_get_gtdb.py
misialq: OK, sounds good, let's leave it 👍

mikerobeson:
Hi @nbokulich,

I noticed the disparity between the taxonomy and sequence counts too. I think the reason is, this is the taxonomy for all of GTDB. That is, not all GTDB genomes have an associated rRNA sequence, and/or or these where removed from the SSU files due to quality issues. I was considering adding a call to `rescript filter-taxa` to remove the extra taxonomy labels in as the extra labels might cause user confusion.

misialq (approved):
Fantastic, thanks @mikerobeson! 🥇

nbokulich:
hey @mikerobeson ,
This also LGTM, thanks! Are you done making changes or is this ready to merge?

> I was considering adding a call to rescript filter-taxa to remove the extra taxonomy labels, as the extra labels might cause user confusion.

No, I do not think that we should do this, leave it to the user to decide.

I think GTDB provides a file of md5 hashes for the downloads, so if you want you could just add an md5 check after the download. Then in case anyone is asking why the files don't match, we can just say (with absolute certainty!) that the action is just downloading exactly what GTDB provides.

mikerobeson:
@nbokulich ahh darn! I just added the "excess taxa filtering" and an associated test. I'll remove them and re-commit. Then it'll be ready to commit.

mikerobeson:
Okay @nbokulich, I removed the excess taxa filtering code. This is ready to merge!

PR #155, opened 2023-04-26
ENH: get_rdp_data

mikerobeson:
Even though there is an avenue for QIIME 2 users to [obtain and import premade RDP files](https://forum.qiime2.org/t/importing-sequence-data-with-lower-case-nucleotide-characters-constructing-an-rdp-classifier-as-an-example/25158) for use in taxonomy assignment. Some users may like to curate the RDP database themselves. This code will provide that ability, while also extracting additional taxonomic rank information not currently available from the premade QIIME 2 compatible RDP files.

The full RDP data is available [here](http://rdp.cme.msu.edu/misc/resources.jsp), and contains the latest information as per the [08/14/2020 RDP release notes](http://rdp.cme.msu.edu/misc/rel10info.jsp).

Currently this draft includes:

- two actions:
  - `parse-rdp-taxonomy`
    - optionally add species labels (*e.g.* like `get-silva-data`)
    - optionally propagate taxonomic ranks
    - choose ranks of interest
  - `get-rdp-data`
    - wraps  `parse-rdp-taxonomy` under the hood
    - downloads both Archaea and Bacteria (16S rRNA gene data) together
    - downloads Fungi (28S rRNA gene data)

**TODO:**
- [ ] add ability to download the Archaea and Bacteria domains individually
- [ ] add ability to download aligned sequences
- [ ] write test code
- [ ] consider refactoring the `get_rdp_data` function and associated code to resemble `get_gtdb_data`

### DIFF
--- a/rescript/get_rdp_data.py
+++ b/rescript/get_rdp_data.py
@@ -1,0 +1,133 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2022-2022, QIIME 2 development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# ----------------------------------------------------------------------------
+
+import os
+import tempfile
+import shutil
+import gzip
+# import warnings
+# import hashlib
+# import pandas as pd
+import qiime2
+from urllib.request import urlretrieve
+# from urllib.error import HTTPError
+# from q2_types.feature_data import MixedCaseDNAFASTAFormat, DNAFASTAFormat
+# from q2_feature_table import merge_seqs, merge_taxa
+
+
+def get_rdp_data(ctx,
+                 ref_db='Archaea+Bacteria',
+                 include_species_labels=False,
+                 rank_propagation=True,
+                 ranks=None):
+
+    parse_taxonomy = ctx.get_action('rescript', 'parse_rdp_taxonomy')
+    merge_rdp_seqs = ctx.get_action('feature_table', 'merge_seqs')
+    merge_rdp_taxonomy = ctx.get_action('feature_table', 'merge_taxa')
+
+    # download data from RDP
+    print('\nDownloading and processing raw files may take some time... '
+          'get some coffee.\n')
+    queries = _assemble_rdp_data_urls(ref_db)
+    rdp_seq_data = _retrieve_data_from_rdp(queries)
+
+    num_seq_files = len(rdp_seq_data)
+
+    if num_seq_files == 2:
+        rdp_tax_df = []
+        rdp_seq_sl = []
+        for rdp_seqs in rdp_seq_data:
+            print('Parsing RDP taxonomy...')
+            taxonomy, = parse_taxonomy(
+                            rdp_reference_sequences=rdp_seqs,
+                            include_species_labels=include_species_labels,
+                            rank_propagation=rank_propagation,
+                            ranks=ranks)
+            # rdp_tax_df.append(taxonomy.view(pd.DataFrame))
+            # rdp_seq_sl.append(rdp_seqs.view(pd.Series))
+            rdp_tax_df.append(taxonomy)
+            rdp_seq_sl.append(rdp_seqs)
+        print('Merging taxonomy ...')
+        merged_rdp_tax, = merge_rdp_taxonomy(data=rdp_tax_df)
+        print('Merging sequences ...')
+        merged_rdp_seqs, = merge_rdp_seqs(data=rdp_seq_sl)
+        print('Writing artifact files...')
+        return merged_rdp_seqs, merged_rdp_tax
+    elif num_seq_files == 1:
+        rdp_seqs = rdp_seq_data[0]
+        rdp_tax, = parse_taxonomy(
+                        rdp_reference_sequences=rdp_seqs,
+                        include_species_labels=include_species_labels,
+                        rank_propagation=rank_propagation,
+                        ranks=ranks)
+        print('Writing artifact files...')
+        return rdp_seqs, rdp_tax
+    elif num_seq_files == 0:
+        raise ValueError('No RDP data to process!')
+    else:
+        raise ValueError('There are more than 2 RDP files! There should '
+                         'be only 1 or 2 files!')
+    # return rdp_seqs, rdp_tax
+
+
+def _assemble_rdp_data_urls(ref_db):
+    '''Generate RDP urls, given database version and reference target.'''
+    # Compile URLs
+    base_url = 'http://rdp.cme.msu.edu/download/'
+
+    if ref_db == 'Archaea+Bacteria':
+        archaea_url = base_url + 'current_Archaea_unaligned.fa.gz'
+        bacteria_url = base_url + 'current_Bacteria_unaligned.fa.gz'
+        return [('archaeal sequences', archaea_url),
+                ('bacterial sequences', bacteria_url)]
+    elif ref_db == 'Fungi':
+        fungi_url = base_url + 'current_Fungi_unaligned.fa.gz'
+        return [('fungal sequences', fungi_url)]
+    else:
+        raise ValueError('Reference database must be either, '
+                         '\'Bacteria+Archaea\' or \'Fungi\'.')
+
+
+def _retrieve_data_from_rdp(queries):
+    '''
+    Download data from RDP, given a list of queries.
+
+    queries: list of tuples of (str, str)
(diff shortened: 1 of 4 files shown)


rescript/get_rdp_data.py
```diff
@@ -0,0 +1,133 @@
...
+# import pandas as pd
+import qiime2
+from urllib.request import urlretrieve
+# from urllib.error import HTTPError
+# from q2_types.feature_data import MixedCaseDNAFASTAFormat, DNAFASTAFormat
+# from q2_feature_table import merge_seqs, merge_taxa
+
+
+def get_rdp_data(ctx,
+                 ref_db='Archaea+Bacteria',
```
nbokulich, line 24: different parameter name? `ref_db` is ambiguous

nbokulich: maybe call this something like "target"?

nbokulich: and the options could then be "16S" and "26S"? (less to write and the markers are then clear from the name)

rescript/get_rdp_data.py
```diff
@@ -0,0 +1,133 @@
...
+
+    # download data from RDP
+    print('\nDownloading and processing raw files may take some time... '
+          'get some coffee.\n')
+    queries = _assemble_rdp_data_urls(ref_db)
+    rdp_seq_data = _retrieve_data_from_rdp(queries)
+
+    num_seq_files = len(rdp_seq_data)
+
+    if num_seq_files == 2:
```
nbokulich, line 41: it does not seem like there is a need to calculate `num_seq_files`, as the only reason you would have > 1 is if `ref_db=Archea+Bacteria`.

I also do not think that you should need to parse the taxonomies separately and then merge. Couldn't you just concatenate the seq and taxonomy files (if ref_db=Archea+Bacteria), then parse taxonomy once?

If you do it that way then I think most of the code within the conditional statement could be removed and reduced to just a few lines.

rescript/get_rdp_data.py
```diff
@@ -0,0 +1,133 @@
...
+                        rdp_reference_sequences=rdp_seqs,
+                        include_species_labels=include_species_labels,
+                        rank_propagation=rank_propagation,
+                        ranks=ranks)
+        print('Writing artifact files...')
+        return rdp_seqs, rdp_tax
+    elif num_seq_files == 0:
+        raise ValueError('No RDP data to process!')
+    else:
+        raise ValueError('There are more than 2 RDP files! There should '
```
nbokulich, line 73: I don't see how you would ever wind up with 0 or >2, seeing as you construct the URLs explicitly. Can these be removed?

rescript/get_rdp_data.py
```diff
@@ -0,0 +1,133 @@
...
+                         'be only 1 or 2 files!')
+    # return rdp_seqs, rdp_tax
```
nbokulich, line 75: ??? this should not be commented out, right?

rescript/get_rdp_data.py
```diff
@@ -0,0 +1,133 @@
...
+            destination = os.path.join(tmpdirname, os.path.basename(query))
+            urlretrieve(query, destination)
+
+            # gunzip on demand (RDP releases are gzipped)
+            try:
+                unzipped_destination = os.path.splitext(destination)[0]
+                print('  Unzipping {0}...'.format(name))
+                _gzip_decompress(destination, unzipped_destination)
+                destination = unzipped_destination
+            except OSError:
```
nbokulich, line 117: what types of errors are encountered here? it seems like you should just error gracefully here if you get an error that would lead to a failed download, instead of failing after attempting to parse the taxonomy later on.

rescript/get_rdp_data.py
```diff
@@ -0,0 +1,133 @@
...
+            print('Importing RDP sequence data...')
+            results.append(qiime2.Artifact.import_data(
+                                                'FeatureData[Sequence]',
+                                                destination,
+                                                'MixedCaseDNAFASTAFormat'))
+        print(results)
+        return results
+
+
+def _gzip_decompress(input_fp, output_fp):
```
nbokulich, line 130: this is already defined here, so should not be replicated:
https://github.com/bokulich-lab/RESCRIPt/blob/026061fb773d5b347a155d0189e6433fb654d804/rescript/get_data.py#L172

rescript/parse_rdp.py
```diff
@@ -0,0 +1,89 @@
+# ----------------------------------------------------------------------------
```
nbokulich, line 1: I recommend tossing this code in the `get_rdp_data.py` file for easier organization (rescript has a lot of different module files at the moment...)

rescript/parse_rdp.py
```diff
@@ -0,0 +1,89 @@
...
+        # parse taxonomy
+        seq_id = seq.metadata['id']
+        sp_label, tax_data = seq.metadata['description'].split(
+                                'Lineage=')
+        new_sp_label = _keep_allowed_chars(
+                                        "_".join(sp_label.strip().split()[:2]))
+        sp_label_dict[seq_id] = new_sp_label
+
+        known_lineage = tax_data.replace('"', '').rsplit(';')
+        if tax_data.endswith(';'):
```
nbokulich, line 41: could't you just use `rstrip(';')` to remove any hanging semicolons instead of checking with this conditional? or am I missing the point?

rescript/parse_rdp.py
```diff
@@ -0,0 +1,89 @@
...
+    sp_label_dict = defaultdict()
+    taxd = defaultdict()
+    for seq in rdp_reference_sequences:
+        # parse taxonomy
+        seq_id = seq.metadata['id']
+        sp_label, tax_data = seq.metadata['description'].split(
+                                'Lineage=')
+        new_sp_label = _keep_allowed_chars(
+                                        "_".join(sp_label.strip().split()[:2]))
+        sp_label_dict[seq_id] = new_sp_label
```
nbokulich, line 38: it's a little confusing that species labels are kept in a separate dict from the rest of the taxonomy labels. Why not just parse and add the species labels directly in the taxonomy dict (if include_species_labels is True)?

wouldn't this also be needed for proper rank_propagation? (i.e., in any cases where the species ID is missing?)

rescript/parse_rdp.py
```diff
@@ -0,0 +1,89 @@
...
+    if ranks is None:
+        ranks = DEFAULT_RDP_RANKS
+
+    # Sort user ranks relative to allowed_ranks
+    # this ensures that the taxonomic ranks supplied by the user
+    # are in order
+    sorted_ranks = [p for r, p in ALLOWED_RDP_RANKS.items()
+                    if r in ranks]
+
+    # If we wish to include species labels we need to add to the sorted_ranks
```
nbokulich, line 74: see comment above; shouldn't/couldn't this be done when compiling the `taxd` dict?

rescript/plugin_setup.py
```diff
@@ -794,6 +797,87 @@
...
+        '(RDP) formatted taxonomy.'
+    },
+    name='Generates an RDP fixed-rank taxonomy.',
+    description=(
+        'Parses FASTA files from the Ribosomal Database Project '
+        '(RDP) reference database to produce a GreenGenes-like fixed rank '
+        'taxonomy. The default generated ranks (and the rank handles used to '
+        'label these ranks in the resulting taxonomy) are: domain (d__), '
+        'phylum (p__), class (c__), subclass (cs_), order (o__), suborder '
+        '(os_), family (f__), and genus (g__). ' + RDP_LICENSE_NOTE
```
nbokulich, line 841: and species?

nbokulich (changes requested):
hey @mikerobeson , great to see this coming together!

I have a few comments inline. I can take a closer look when the unit tests are ready.

One overarching comment/question: why separate `get_rdp_data` and `parse_rdp_taxonomy`? It seems like these could/should just be a single action.

mikerobeson:
I was [notified](https://forum.qiime2.org/t/using-gtdb-to-classify-species-level/27109/5) that there was a discontinuation of funding for RDP. I followed up with this user privately, and this does seem to be the case.  This also explains why the [http://rdp.cme.msu.edu/](http://rdp.cme.msu.edu/) website has been down. It does appear the associated [sourceforge](https://sourceforge.net/projects/rdp-classifier/) and [GitHub](https://github.com/rdpstaff) sites will remain.

For this reason I'll be closing this PR. :-(

nbokulich:
ah too bad! thanks @mikerobeson !

PR #162, opened 2023-09-28
Update ci-dev.yaml

ebolyen:
We missed the fact that this has a dependency on q2-types-genomics, so putting it in the shotgun distribution for the moment.

Also, would it be possible to request push access to this repo?

(I need the a manual trigger of [this action](https://github.com/bokulich-lab/RESCRIPt/actions/workflows/join-release.yaml) to run to get this into the 2023.9 release).

### DIFF
--- a/.github/workflows/ci-dev.yaml
+++ b/.github/workflows/ci-dev.yaml
@@ -9,4 +9,4 @@ jobs:
   ci:
     uses: qiime2/distributions/.github/workflows/lib-ci-dev.yaml@dev
     with:
-      distro: core
+      distro: shotgun


nbokulich:
can q2-types-genomics also go in core then? Most of the actions in RESCRIPt are for amplicon databases so it would be a little silly to put it in which shotgun (where it has a use, but less) but not in the amplicon/core.

@misialq could you please give @ebolyen push access?

misialq:
Hey @ebolyen, you should now have access 🚀

ebolyen:
Thanks! Yeah that might be an inevitable conclusion, are you ok with us just figuring out something that will work and then we can shuffle things around for the next cycle to make more sense?

(cc'ing @gregcaporaso as he mentioned we might need to do something like the above)

nbokulich:
@ebolyen yeah of course! Makes sense.

ebolyen:
Ok, we're back to green! We'll figure out a plan for moving this back to amplicon for the next cycle.

Issue #163, opened 2023-09-28
Remove heavy dependencies


ebolyen:
It looks like we may be able to drop some rather significant weight from the recipe:

```yaml
    - q2-longitudinal {{ qiime2_epoch }}.* 
    - q2-feature-classifier {{ qiime2_epoch }}.*
```
I can't find an import of longitudinal, which is bringing along sample-classifier as well.
And for feature-classifier, the only imports are [these](https://github.com/bokulich-lab/RESCRIPt/blob/11aedaa7dc46a3e4f0b9523b9ff4043f60e5a999/rescript/plugin_setup.py#L40-L42) which I don't think is actually worth the cost of pulling feature-classifier along (requiring quality-control as well).

nbokulich:
Yeah the longitudinal dependency can definitely be dropped... I think that is historical.

dropping q2-feature-classifier would require dropping a couple actions in RESCRIPt — is that what you are proposing? Or add instructions that users who wish to run those actions will need to install q2-feature-classifier separately?

ebolyen:
No, I would want to keep those actions, but I think we can just duplicate the parameter descriptions (and move the type over to q2-types) (from these imports here: https://github.com/bokulich-lab/RESCRIPt/blob/11aedaa7dc46a3e4f0b9523b9ff4043f60e5a999/rescript/plugin_setup.py#L40-L42).

Is there a pipeline that uses feature-classifier directly that I am missing?

nbokulich:
> Is there a pipeline that uses feature-classifier directly that I am missing?

Yes, see e.g.:
https://github.com/bokulich-lab/RESCRIPt/blob/11aedaa7dc46a3e4f0b9523b9ff4043f60e5a999/rescript/cross_validate.py#L38-L39

I would be very tempted to deprecated these actions though. Maybe just port those pipelines over into the classifier training workflows to use them for our purposes. I was never very happy with those pipelines.

lizgehret:
This seems like a good candidate to finish up in this upcoming release cycle since we are adding RESCRIPt to the amplicon distro. @nbokulich I can bring this up in this week's eng meeting - your preference would still be to deprecate these actions?

nbokulich:
Hi @lizgehret thanks for the reminder! Yes let's discuss this week.

Which dependencies are priority to drop? It looks like q2-longitudinal is indeed used in a couple places, e.g., here:
[...]

PR #164, opened 2023-10-02
DEP: drop q2-longitudinal

ebolyen:
Turns out this is causing some issues with conda and matplotlib that don't really _need_ to be an issue right now.

https://github.com/qiime2/distributions/actions/runs/6385353574/job/17331623684?pr=27

This isn't an issue with the amplicon distro for whatever reason, but we haven't looked into it too closely yet.

Related to #163


ebolyen:
I suppose CI will fail without q2-types-genomics...

ebolyen:
So it did build the package, which I'm going to call a sufficient test for a recipe.

ebolyen:
Sorry for the self-merge, and we appreciate ya'lls patience as we fix up our CI to work correctly.

PR #134, opened 2022-04-10
get_unite_data functions

colinbrislawn:
Early code to close https://github.com/bokulich-lab/RESCRIPt/issues/123

PR #166, opened 2023-11-14
patch get_unite

colinbrislawn:
Looks like the merged version of #134 was pretty old.

I have no idea how that happened, but thanks to git, I've got all the code.

Is this a hot patch? Let me know how to proceed...


rescript/get_unite.py
```diff
@@ -10,10 +10,13 @@
...
 import requests
-from requests.exceptions import HTTPError
 
 from pandas import DataFrame
-from q2_types.feature_data import TaxonomyFormat, DNAFASTAFormat, DNAIterator
+from q2_types.feature_data import (
+    TaxonomyFormat,
+    MixedCaseDNAFASTAFormat,
+    DNAIterator,
+)
```
colinbrislawn, line 19: @mikerobeson This would explain why your testing showed no slowdown!

colinbrislawn:
Black and pytest pass. I think this is ready to go!

rescript/get_unite.py
mikerobeson: Haha right! I'll do a quick download and run. If that all works I'll just go ahead and merge.

PR #174, opened 2024-02-06
ENH: adds `--p-n-threads` option to `trim-alignment`

### LINKED ISSUE #127: `trim-alignment` : expose multithreading
`mafft` has a `--threads` option that is exposed in q2-alignment's `mafft-add` action. But this is not exposed in RESCRIPt. This option should be exposed to enable multithreading. (or maybe this is intended? when adding primers to a larger alignment does multithreading offer any advantage?)

mikerobeson:
Fixes #127 

Also, made some minor help text and documentation edits.

### DIFF
--- a/rescript/trim_alignment.py
+++ b/rescript/trim_alignment.py
@@ -190,11 +190,13 @@ def _trim_alignment(expand_alignment_action,
                     primer_fwd=None,
                     primer_rev=None,
                     position_start=None,
-                    position_end=None) -> AlignedDNAFASTAFormat:
+                    position_end=None,
+                    n_threads=1) -> AlignedDNAFASTAFormat:
     """
     Trim alignment based on primer alignment or explicitly specified
     positions. When at least one primer sequence is given, primer-based
-    trimming will be performed, otherwise position-based trimming is done.
+    trimming will be performed via the `addfragments` option of mafft,
+    otherwise position-based trimming is done.
 
     Arguments:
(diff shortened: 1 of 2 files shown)


rescript/trim_alignment.py
```diff
@@ -190,11 +190,13 @@ def _trim_alignment(expand_alignment_action,
...
                     position_start=None,
-                    position_end=None) -> AlignedDNAFASTAFormat:
+                    position_end=None,
+                    n_threads=1) -> AlignedDNAFASTAFormat:
     """
     Trim alignment based on primer alignment or explicitly specified
     positions. When at least one primer sequence is given, primer-based
-    trimming will be performed, otherwise position-based trimming is done.
+    trimming will be performed via the `addfragments` option of mafft,
+    otherwise position-based trimming is done.
```
nbokulich, line 199: the `n_threads` param would not be effective as far as I can tell when position-based trimming is done, so this should be mentioned in the param description

nbokulich (changes requested):
thanks @mikerobeson ! Small comment inline. Have you tested the performance speed-up with multi-threading? I can take a look, but curious how much of a bump we see.

rescript/trim_alignment.py
mikerobeson: Oooh, good catch!

nbokulich:
Hey @mikerobeson ,

LGTM, thanks!

Issue #183, opened 2024-04-25
ENH: Add the ability to fetch the latest GTDB database (v220) to `get-gtdb-data`


mikerobeson:
GTDB has just released an updated version of their database [v220](https://data.gtdb.ecogenomic.org/releases/release220/220.0/). Detailed information about this release is [here](https://gtdb.ecogenomic.org/stats/r220). We should add the appropriate versions to `get_dtdb.py`, by adding to this [dictionary](https://github.com/bokulich-lab/RESCRIPt/blob/6e87cfeb239c7cfb91126a6ed5a2e64622343bde/rescript/get_gtdb.py#L25), and set the default [here](https://github.com/bokulich-lab/RESCRIPt/blob/6e87cfeb239c7cfb91126a6ed5a2e64622343bde/rescript/get_gtdb.py#L32). Then update `plugin_setup.py` [here](https://github.com/bokulich-lab/RESCRIPt/blob/6e87cfeb239c7cfb91126a6ed5a2e64622343bde/rescript/plugin_setup.py#L956) 

Sadly it looks like they changed some of the `tar.gz` files to `fna.gz`. 🤨

nbokulich:
fixed by #184

PR #184, opened 2024-04-25
ENH: Adds ability to fetch GTDB v220 to get-gtdb-data

mikerobeson:
Resolves issue #183.

Also includes a refactor of some code... That is, GTDB v220 now hosts the FASTA files as `fna.gz`, while files for the older versions of the database continue to be hosted as `tar.gz`.

### DIFF
--- a/rescript/get_gtdb.py
+++ b/rescript/get_gtdb.py
@@ -8,7 +8,9 @@
 
 import os
 import tempfile
+import shutil
 import tarfile
+import gzip
 import warnings
 import pandas as pd
 from urllib.request import urlretrieve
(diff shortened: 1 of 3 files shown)


rescript/get_gtdb.py
```diff
@@ -52,6 +55,12 @@ def _assemble_queries(version='214.1',
     # ^^ Set `base_version` variable becuase number after the decimal is
     # only used for the directory. GTDB trims this off for the actual
     # file names...
+    # GTDB v220 started storing the ssu_reps FASTA files
+    # as 'fna.gz' instead of their usual 'tar.gz'.
+    if version == '220.0':
```
nbokulich, line 60: should this be >=? Or do you want to wait to react until the next release is out in case they change again?

nbokulich (changes requested):
thanks @mikerobeson ! I have one small comment inline. Also, the tests appear to be failing. Once these are passing (and you address my comment) I think this will be good to merge, thanks!

rescript/get_gtdb.py
nbokulich: never mind, I realize that this is evaluating a string, so ignore my comment. We can only address this when we have a new version (though we might want to instead check if it is NOT in the list of older versions, otherwise we need to update the list each time a new version is released).

PR #186, opened 2024-05-29
CI: adds command to run tests

lizgehret:
this needs to be added so that the unit test suite actually runs for all CI actions (ci-dev, prepare, release, etc)


ebolyen:
Looks like we need to update the target distro in ci.yml as well.

PR #188, opened 2024-06-04
TST: fixing test failures from python 3.9 update

lizgehret:


lizgehret:
merging b/c this needs to be implemented in order for for an amplicon prepare to succeed

PR #189, opened 2024-06-07
TEMPFIX: skip unit tests that rely on q2-longitudinal if it is not installed

nbokulich:
This is only temporary, until we finish replacing the q2-longitudinal dependency with q2-visard.

We will skip tests that rely on q2-longitudinal if q2-longitudinal is not installed in the environment.

### DIFF
--- a/rescript/tests/test_cross_validate.py
+++ b/rescript/tests/test_cross_validate.py
@@ -21,7 +21,7 @@ from rescript import cross_validate
 import_data = qiime2.Artifact.import_data
 
 try:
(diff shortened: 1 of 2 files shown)


misialq (commented):
While I'm running the test, something else that occurred to me now: should this PR also disable the action registration in the same case? The current changes only allow skipping the tests but technically an unlucky user could still use the action that relies on q2-longitudinal?

nbokulich:
closing in favor of #190

PR #191, opened 2024-06-07
BUG: gtdb taxonomy parser

nbokulich:
Evidently #169 introduced a bug into the way taxonomy is parsed from GTDB files. Taxonomy labels were being split on whitespace, so only the genus name was listed under the species rank.

This fixes this issue by also including the species label, and introduces a simple test for the parser, just in case the taxonomy format changes in the future.

I manually tested and confirmed that this works with versions 202, 214, 220. The taxonomy format is consistent — see the test for the structure.

FYI @mikerobeson


nbokulich:
is it worth patching the latest release? @lizgehret ?

PR #193, opened 2024-06-11
FIX: update `get-ncbi-genomes` to use genome accession IDs as taxonomy IDs

misialq:
This PR fixes the issue where the taxonomies generated by the `get-ncbi-genomes` action would contain IDs that do not match the IDs used in the corresponding FASTA file to represent genomic sequences. 

- Previously, genome _assembly_ IDs (GCF*) were used in the taxonomy but genome _accession_ numbers (e.g., NC*) were used in sequence headers obtained from NCBI. 
- The changes introduced here will allow finding the accession ID in the sequence metadata obtained with the sequences and use this one instead of the assembly ID.
- An additional challenge which is solved by this approach is the potential existence of _multiple_ chromosomes in a single assembly (hence, single ID would be have been used to represent multiple sequences): in case of bacterial, the typical scenario is existence of additional plasmid molecules; in eukaryotes it may be gDNA molecule comprising multiple chromosomes + mitochondrial/chloroplast DNA
- Additionally, a new flag is being introduced: `--p-only-genomic` which will allow users to only keep the genomic DNA vs. keeping all the molecules.


nbokulich (changes requested):
Hi @misialq ,

Looks good. Could you please remove the `ncbi-datasets.zip` and `ncbi-dataset/README.md` files if they are not needed?

I still want to pull down the code and test it out with a few different queries, but in the meantime I just have one small comment on the code.

misialq:
Hey @nbokulich, thanks for review! I removed the README file - the zip one had already been removed (you meant the one in the tests/data dir, right?)

nbokulich:
Hey @misialq I did a bit of user testing with my favorite taxa and all LGTM, thanks!

Yes you are right, I was not reading carefully enough. I now see that this PR removes the unwanted file.

misialq:
Hey @nbokulich, thanks a lot for testing! In that case, I'm merging this with great excitement 😆

Issue #195, opened 2024-07-10
ENH: extract-seq-segments add param "min_coverage"


nbokulich:
# The Problem
`extract-seq-segments` does not have a `min_coverage` threshold exposed, so it is possible to extract sequences that may pass the %identity threshold but actually have very low coverage (I do not know off-hand what the default is, but we should set an explicit default either way).

# Solution
Expose a `min_coverage` parameter to specify the minimum amount of coverage of the reference that is required for a hit to pass.

This code should be relatively easy to expose — there is precedent with the `vsearch-global` action in q2-feature-classifier that can be used as a template here. The one difference (I think) is that in q2-feature-classifier the `min_coverage` parameter is about min coverage of the query; whereas in `extract-seq-segments` this should be the min coverage of the reference, which is the shorter sequence in this case that is used to recruit additional sequences.

nbokulich:
resolved by #199

PR #199, opened 2024-08-06
ENH: Added ability to specify segment coverage value for `extract-seq-segments`.

mikerobeson:
Adresses #195 

This PR adds the option `--p-target-coverage` to `extract-seq-segments`. This denotes the amount of coverage the sequence segments must have in order to extract the corresponding match from a longer sequence.


rescript/plugin_setup.py
```diff
@@ -1141,6 +1144,8 @@
     parameter_descriptions={
                 'threads': VSEARCH_PARAM_DESCRIPTIONS['threads'],
                 'perc_identity': VSEARCH_PARAM_DESCRIPTIONS['perc_identity'],
+                'target_coverage': 'Minimum coverage of reference segment '
```
nbokulich, line 1147: This param description is a little confusing. Could you please make it clearer?

Also since this is `target_coverage`, shouldn't this be the minimum coverage of the target, which is the reference sequence segments in this case? As written now it sounds like this is coverage of the input sequences.

nbokulich (changes requested):
Hey @mikerobeson looks good to me, I just have a minor comment inline. Thanks!

rescript/plugin_setup.py
mikerobeson: I've updated the help text to be more clear.

nbokulich (approved):
thanks @mikerobeson ! Looks great

PR #203, opened 2024-09-04
ENH: Adds NCBI api-key option

### LINKED ISSUE #145: Use NCBI API keys in get-ncbi-data command
It could be useful when querying NCBI using rescript to be able to provide the user's [API key](https://ncbiinsights.ncbi.nlm.nih.gov/2017/11/02/new-api-keys-for-the-e-utilities/), to increase the number of requests per second allowed.

mikerobeson:
Attempts to fix  #145, by adding the `--p-api-key` option to both `get-ncbi-data` and `get-ncbi-data-protein`. It does this by appending `api_key` to the `_entrez_params` dictionary so that it is available for all NCBI requests.

To do:

- Test code
  - I could use help on the best way to write test code for this...
  - Might test reading in a mock api-key?

### DIFF
--- a/rescript/ncbi.py
+++ b/rescript/ncbi.py
@@ -82,7 +82,7 @@ def get_ncbi_data(
         raise ValueError('Query or accession_ids must be supplied')
     if api_key:
         try:
-            (api_key,) = api_key.get_ids()
+            api_key, = api_key.get_ids()
             _entrez_params['api_key'] = api_key
             global _entrez_delay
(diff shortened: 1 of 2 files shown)


mikerobeson:
Going to close this PR as it is clear that NCBI requests are not the rate-limiting step. At least for the purposes being used here. This was confirmed after several rounds of testing. There was no difference in overall run time when downloading marker gene data from NCBI, with or without the `api-key`.

Issue #194, opened 2024-06-24
get-ncbi-genomes: custom rank information


nbokulich:
As a user, I would like to occasionally customize the taxonomic rank information provided for the genomes that I download, e.g., to include domain or subkingdom or even infraclass. 

Right now, this is not possible and the retrieved ranks are the standard/bacterial ranks (a fine default).

I would like to be able to specify which ranks I want to retrieve, e.g., similar to the action `get-ncbi-data`.

mikerobeson:
fixed by #207

PR #207, opened 2024-10-23
ENH: Added ranks and rank-propagation parameters to `get-ncbi-genomes`

VinzentRisch:
solves #194 

- Added ranks and rank-propagation parameters to `get-ncbi-genomes`.
- The parameters behave the same as in the `get-ncbi-data` function

### DIFF
--- a/rescript/ncbi_datasets.py
+++ b/rescript/ncbi_datasets.py
@@ -156,12 +156,14 @@ def _fetch_and_extract_dataset(
 def _fetch_taxonomy(
         all_acc_ids: list,
         all_tax_ids: list,
-        accession_to_assembly: pd.Series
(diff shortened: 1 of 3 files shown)


nbokulich:
hey @mikerobeson just to confirm since I did not see any review comments before you merged: everything checks out with rank propagation and custom ranks?

mikerobeson:
Hi @nbokulich, I entered in my comments in the merge dialog... I guess that did not take? But yes, rank propagation worked as well as selecting different rank options. I also checked the lint and test cases.. all passed.

nbokulich:
hey @mikerobeson no it looks like these were hidden. Thanks for testing and confirming!

PR #208, opened 2024-11-02
ENH: adds `keeplength` option to ensure the correct region is extracted when using primers.

### LINKED ISSUE #175: providing primers for `trim-alignment`  provides different output vs explicit column positions for the same primers.
While working on the new `qiime rescript trim-alignment` tutorial, I started to realize that the output of `trim-alignment` does not seem to return expected results when using primer sequences for extracting the amplicon region, vs providing the amplicon positions directly.

When I run `qiime alignment mafft-add --p-addfragments --i-sequences my_primers.qza ...` and look at the exported output, I do observe what I expect. That is, I see the primer aligned where it should, and it maps to the explicit positions. In fact, I used this approach to help me map to these positions. However, when I run `qiime rescript trim-alignment...` the resulting extracted data appears spurious. It appears, in some cases that the sequences change / are different in the output, or the wrong region is extracted. Also, if I use 10 sequences vs 5k sequences, very different regions are extracted when using the primer search option. I am still not sure if the issue is myself, or not, so...

Anyway, I am still investigating this, and will try and come up with a better set of test cases, but perhaps, others can help me look into this. I suggest pulling the SILVA alignment, and subsample it for testing. Then compare the `qiime alignment mafft-add --p-addfragments` output with the `qiime rescript trim-alignment` outputs. I've been using the Unipro UGENE software to quickly visualize the SILVA alignments with the added primer sequences, etc.

**Extra info and example commands**

**V4V5	primer info:**
515F (Parada) - GTGYCAGCMGCCGCGGTAA	
926R (Quince) - CCGYCAATTYMTTTRAGTTT

Expected alignment positions *including* the primer region from 50k alignment:
11895 - 28464

Expected alignment positions *excluding* the primer region region from 50k alignment:
13862 - 27655

Running the following will result in a combined alignment with several thousand positions less than there should be. That is: 43,515 alignment positions instead of 50,000. The problem is that the calculated position range, for the above "excluding the primer region" changes from 13862 - 27655 (50k alignment) to 13862 - 21844 (43k alignment). 

```
qiime alignment mafft-add \
    --i-alignment silva-138-1-nr99-aln-dna-ss01.qza \
    --i-sequences V4V5_515F_926R_primers.qza \
    --p-addfragments \
    --o-expanded-alignment silva-138-1-nr99-aln-dna-ss01-v4v5.qza
```
Can confirm this by exporting to FASTA and view in your favorite alignment editor. 

The following command , *i.e.* maintains the 50k alignment positions, and returns the proper alignment ranges. If you re-run the command without `--keeplength` you should obtain the same result as the `qiime alignment mafft-add` command.

```
mafft --preservecase --inputorder --thread 8 --keeplength   \
    --addfragments V4V5_515F_926R_primers.fasta  \
    silva-138-1-nr99-aln-dna-ss01-export/aligned-dna-sequences.fasta   \
    > aligned-dna-sequences-v4v5-pq.fasta
```

Compare to the outputs of:
```
qiime rescript trim-alignment \
    --i-aligned-sequences silva-138-1-nr99-aln-dna-ss01.qza \
    --p-primer-fwd GTGYCAGCMGCCGCGGTAA \
    --p-primer-rev CCGYCAATTYMTTTRAGTTT \
    --o-trimmed-sequences silva-138-1-nr99-aln-dna-ss01-v4v5-primer-trimmed.qza
```

and 

```
qiime rescript trim-alignment \
    --i-aligned-sequences silva-138-1-nr99-aln-dna-ss01.qza \
    --p-position-start 11895 \
    --p-position-end 28464 \
    --o-trimmed-sequences silva-138-1-nr99-aln-dna-ss01-v4v5-pos-trimmed.qza
```

**EDIT:**
It appears that the issue can be partially resolved by adding mafft's [--keeplength parameter](https://mafft.cbrc.jp/alignment/software/addsequences.html) to `qiime alignment mafft-add` for our specific purpose here. Otherwise the final combined alignment length can change. Which is what is happening in my case. That is the alignment is shrinking throwing off alignment positions when using `mafft-add`. I've opened an issue [here](https://github.com/qiime2/q2-alignment/issues/83).

mikerobeson:
Fixes  #175. This is now possible as the `keeplength` option was added to the  [q2-alignment plugin](https://github.com/qiime2/q2-alignment/).

Adds `keeplength` parameter to make sure original alignment length does not change. Otherwise there is potential for the incorrect region to be extracted from the alignment. This is important to make sure alignment positions can easily be referenced to the original alignment. See the [q2-alignment plugin](https://github.com/qiime2/q2-alignment/), specifically the [mafft-add test cases](https://github.com/qiime2/q2-alignment/blob/dev/q2_alignment/tests/test_mafft.py) for examples.

Also adds clearer clarification and a warning within the `trim-alignment` description when using primers to find and extract amplicon regions.

All of the original `trim-alignment` tests still pass.


rescript/trim_alignment.py
```diff
@@ -221,10 +221,17 @@ def _trim_alignment(expand_alignment_action,
...
+        # does not change. Otherwise there is potential for the incorrect
+        # region to be extracted from the alignment. This is important
+        # to make sure alignment positions can easily be referenced
+        # to the original alignment. See the 'q2-alignment' plugin,
+        # specifically the 'mafft-add' test cases for examples.
         alignment_with_primers, = expand_alignment_action(
             alignment=aligned_sequences,
             sequences=primers,
             addfragments=True,
+            keeplength=True,
```
nbokulich, line 234: hey @mikerobeson do we want to hardcode this param? Should we make it True by default but expose the option to the user? Is there ever a case when they might want to not keep the length? In this case this would only happen if there are indels in the primer region, right? which we probably don't want to preserve.

nbokulich (changes requested):
Looks good, thanks @mikerobeson ! I have one small question inline. 

Also: could you maybe add a tiny unit test to make sure this is working as intended? Maybe you have a test case where the primer introduced gaps in an alignment and led to incorrect results? Or is it all just very theoretical?

rescript/trim_alignment.py
mikerobeson: These primer indel cases are explicitly tested in the [mafft-add](https://github.com/qiime2/q2-alignment/blob/dev/q2_alignment/tests/test_mafft.py) tests by toggling that option on/off. I was thinking I did not need to reproduce that test case here. 

The issue is that the final alignment length changes after the primer sequences are added. Often, due to common column gaps in the newly combined alignment being removed. From what I can observe this is regardless if new indels are added within the location of the newly aligned primers. If any of the V3V4, V4, V4V5 primers are used no indels are added to the location where the primers align.

Due to the alignment length change after the primers are added to the alignment, the derived positional values of the primers will differ from the proper positions of the original alignment. Thus, when these derived positional values are used, the actual region extracted is not the region we are after, as they are based on the positions of the new alignment.

I suppose I could add another set of small files based on a few of the actual SILVA  alignment sequences. Then I can check the expected vs observed returned alignment length differences. I'd just have to override  `keeplength=True` option and compare.

mikerobeson: This issue is the trimming aspect... we are deriving the positional values from the resulting alignment after the primers are added. Then using those values to trim from the original alignment. This won't work in all cases as the resulting alignment length can very likely change. Which is why we need to use `keeplength=True`. I think that toggling this option makes obvious sense for the general `mafft-add` function... but not for our specific use case here: finding alignment position trimming locations.

nbokulich: okay now I understand! So any length change messes up the positional trimming. I agree, hardcoding sounds necessary in that case.

And agreed, unit tests are not needed if this is explicitly tested in q2-alignment.

mikerobeson: I just realized that I can likely fix this by adding an `if` statement to check if primers were used. If primers are used then this [line](https://github.com/bokulich-lab/RESCRIPt/blob/49fc648043af56dfafa41a030ac4927bab4f4a83/rescript/trim_alignment.py#L240) could be something like: 

```
result = _trim_all_sequences(aligned_sequences_with_primers, trim_positions)
```

Then the primer sequences can removed from that output prior to saving. This would enable us toggle the `keeplength` option. For instance if users would like to concatenate their edits to some master alignment and not worry about alignment length changes. 🤔

mikerobeson: Hi @nbokulich, I've added explicit unit test for the keep length option. I decided to forgoe toggling the `--p-keeplengh` parameter, as it is expected that this be used when trying to find the location of added fragments. In fact, the mafft parameter `--mapout` (currently not accessible within the plugin) enforces the mafft `--keeplength` function if it is not supplied. Thus,, I think we are in keeping with the spirit of the tool when it comes to determining alignment positions. :-)

mikerobeson: Actually, I'm still working on the test cases... I'll let you know when I am done. :-)

mikerobeson: Okay... I realized that I should actually run my test cases rather than mock them. Should be read to review now. :-)

mikerobeson:
Appears to be failing on `ModuleNotFoundError: No module named 'qiime2.plugins.alignment'`. Not sure why.... When I run the tests locally within the `q2dev-amplicon` environment all the tests pass. This applies for both the `2024.10` and `2025.4` versions of `dev`.

nbokulich:
looks like q2-alignment is missing from the recipe. Try adding this and re-run?

mikerobeson:
Well, that's embarrassing. That is what the issue was... 🤦

Everything passes. 🎉

nbokulich:
thanks for the ping @mikerobeson looks good!

PR #209, opened 2024-11-04
ENH: added `--p-max-seq-len` to `extract-seq-segments` and associated test code

mikerobeson:
This PR adds to `qiime rescript extract-seq-segments`:

- `--p-max-seq-len`: to prevent long reference segment sequences to be used for searching.
- `test_extract_seq_segments_min_max`: tests setting `--min-seq-len` & `--max-seq-len` reference segment sequences.


rescript/extract_seq_segments.py
```diff
@@ -16,6 +16,7 @@ def extract_seq_segments(input_sequences: DNAFASTAFormat,
                          perc_identity: float = 0.7,
                          target_coverage: float = 0.9,
                          min_seq_len: int = None,
+                         max_seq_len: int = None,
```
VinzentRisch, line 19: I think it would be better if the parameters min_seq_len and max_seq_len where not set to None per default but rather directly to the vsearch default parameters of 32 and 5000 (if I looked correctly in the documentation). 
I think this would be more clear and a user would not have to check in the vsearch documentation for the parameters. 
Or is there a reason why you set them to None? 
```suggestion
                         min_seq_len: int = 32,
                         max_seq_len: int = 5000,
```
```
  --minseqlength INT          min seq length (clust/derep/search: 32, other:1)
  --maxseqlength INT          maximum sequence length (50000)
```

rescript/plugin_setup.py
```diff
@@ -1151,10 +1152,14 @@
...
-                               'for searching. Any sequence less than '
+                'min_seq_len': 'Minimum length of reference sequence segment '
+                               'allowed for searching. Any sequence less than '
                                'this will be discarded. If not set, default '
-                               'program settings will be used.'},
+                               'program settings will be used.',
+                'max_seq_len': 'Maximum length of reference sequence segment '
+                               'allowed for searching. Any sequence greater '
+                               'than this will be discarded. If not set, '
+                               'default program settings will be used.'},
```
VinzentRisch, line 1162:
```suggestion
                               'than this will be discarded.'},
```
This could be removed with the change from above.

rescript/extract_seq_segments.py
```diff
@@ -49,5 +50,7 @@ def extract_seq_segments(input_sequences: DNAFASTAFormat,
                '--notmatched', no_match_f.name]
         if min_seq_len:
             cmd += ['--minseqlength', '%i' % min_seq_len]
+        if max_seq_len:
+            cmd += ['--maxseqlength', '%i' % max_seq_len]
```
VinzentRisch, line 54: Also these if statements could be removed and the parameters could be passed with all the others with the change above.

VinzentRisch (changes requested):
Hi @mikerobeson 
Everything looks good to me, except for one suggestion below.

rescript/extract_seq_segments.py
mikerobeson: Historically we've avoided hard setting program defaults. The main reason is that there have been cases in the past where the developer changed the default settings, and the QIIME package was not updated... leading to differences in the output processed by QIIME compared to the original tool. This requires us to be more vigilant in updating our defaults. This is why many options in several QIIIME 2 plugins state that "default program settings will be used". This way we do not have to worry about parameter changes. I'd rather like to avoid setting these. I can mention these common default settings as recommendations in the help text.

mikerobeson:
Hi @VinzentRisch, I've incorporated your suggestions. Thank you so much for the review! :-)

VinzentRisch (approved):
Yes great work! feel free to merge after you resolved the conflicts :)

mikerobeson:
Merging and this PR based on comments from @VinzentRisch. 🎉

PR #218, opened 2025-03-31
MAINT: migrate get-ncbi-genomes action to use the updated NCBI API

misialq:
This PR migrates the NCBI Datasets Python client within the `get-ncbi-genomes` action to its newer version which supports the v2 NCBI API (see [here](https://ncbiinsights.ncbi.nlm.nih.gov/2024/02/20/updates-ncbi-datasets-apis-command-line-tools/) for details).

**Note**: since the new version supports fetching genomes for multiple taxa within a single query, the `taxon` parameter got renamed to `taxa` and now accepts a list of taxon names or IDs.

### DIFF
--- a/rescript/ncbi_datasets.py
+++ b/rescript/ncbi_datasets.py
@@ -15,12 +15,11 @@ import zipfile
 from copy import deepcopy
 from typing import List, Dict
 
-import ncbi.datasets as nd
-from ncbi.datasets.openapi import ApiClient as DatasetsApiClient
+from ncbi.datasets.openapi import ApiClient as DatasetsApiClient, GenomeApi
 import pandas as pd
 import skbio
 from multiprocessing import Manager
-from ncbi.datasets import ApiException
+from ncbi.datasets.openapi.rest import ApiException
 from q2_types.feature_data import DNAFASTAFormat
 from q2_types.genome_data import (LociDirectoryFormat,
                                   ProteinsDirectoryFormat)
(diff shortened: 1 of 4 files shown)


lizgehret:
testing these changes via a Lang- PR in distributions: https://github.com/qiime2/distributions/pull/668

as long as this metapackage resolves and test suites pass, this will be good to merge!

rescript/ncbi_datasets.py
```diff
@@ -30,14 +29,14 @@
 
 def _get_assembly_descriptors(
         api_instance, assembly_levels, assembly_source, only_reference,
-        page_size, taxon, tax_exact_match
+        page_size, taxons, tax_exact_match
```
VinzentRisch, line 32:
```suggestion
        page_size, taxa, tax_exact_match
```
Why are you using sometimes the word taxons and sometimes taxa? Taxons is not correct or is there a reason?

VinzentRisch (changes requested):
I ran it and it works without problems. Just one suggestion below.

rescript/ncbi_datasets.py
misialq: That's because the NCBI API uses `taxons`. But as you, correctly, pointed out that's not quite right so in my own code I'm using `taxa` instead (but perhaps not quite consistently) - I now changed it everywhere where not absolutely required.

VinzentRisch (approved):
Looks good to me!

Issue #216, opened 2025-02-27
update UNITE to 2025-02-19


colinbrislawn:
https://unite.ut.ee/repository.php

nbokulich:
Let's add this this week so it makes it in the 2025.4 release.

We already have version 10.0 in RESCRIPt, but confusingly UNITE releases multiple versions under the same version number starting in version 9.0. @colinbrislawn @mikerobeson do you know why this is? I suggest that we move to using the release date instead of the "version number" for UNITE releases in RESCRIPt. In the UNITE release notes they refer to this as, e.g., "Version 19.02.2025" so it feels justified to just use the release date as the version number in RESCRIPt from now on. Any objections?

colinbrislawn:
Looks good to me!

I like that we are matching the release data [from the website](https://unite.ut.ee/repository.php), and not from the filename as that's what users are more likely to see.

I wanted to ask about that, because sometimes those dates don't match!
The main download page says 2023-07-18 (first published), the file says 2023-07-25 (last published). 
![Image](https://github.com/user-attachments/assets/848fc713-ee5a-4d9e-9d3b-41c7a0d20884)

Using dates would also let us provide 9.0 old (2022-10-16) and 9.0 new (2023-07-18), which we could not do before.

mikerobeson:
Yeah I decided that users are going to more likely see the dates as listed on the web site rather than the file names. I figured it'd be less confusing.

nbokulich:
Closing as this was fixed in #219 

but please re-open if I'm missing something!

PR #219, opened 2025-04-01
ENH: updating version number with release date

mikerobeson:
Addresses #216.

Replaces version, *e.g.* `10.0` with the latest release date `2025-02-19` (and doi) for that version.

### DIFF
--- a/rescript/get_unite.py
+++ b/rescript/get_unite.py
@@ -20,17 +20,18 @@ from q2_types.feature_data import (
 
 # Source: https://unite.ut.ee/repository.php
 UNITE_DOIS = {
-    "10.0": {
+    # Old version 10.0 is not listed here
+    "2025-02-19": {
         "fungi": {
-            False: "10.15156/BIO/2959336",
-            True: "10.15156/BIO/2959337",
+            False: "10.15156/BIO/3301241",
+            True: "10.15156/BIO/3301242",
         },
         "eukaryotes": {
-            False: "10.15156/BIO/2959338",
-            True: "10.15156/BIO/2959339",
+            False: "10.15156/BIO/3301243",
+            True: "10.15156/BIO/3301244",
         },
     },
-    "9.0": {
+    "2023-07-18": {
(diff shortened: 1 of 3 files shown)


colinbrislawn:
Can we keep 2024-04-04 and simply add the new 2025-02-19 release?

mikerobeson:
Done. Added both `2024-04-04` and `2022-10-16` to the list.

nbokulich (changes requested):
Thanks @mikerobeson ! My one sneaking worry is if this will introduce breaking changes in anyone's pipelines/existing code. What if we keep the old numbered versions as aliases of the dated releases? Is that getting too nitpicky/messy? I suppose folks probably are not making these on the fly as parts of pipelines so maybe we are okay 🤷

mikerobeson:
Hi @nbokulich, I do prefer the dates, mainly for the reasons that @colinbrislawn brought up:
>Using dates would also let us provide 9.0 old (2022-10-16) and 9.0 new (2023-07-18), which we could not do before.

I suppose we could keep some as aliases too. I was thinking of doing something like this after making the `UNITE_DOIS` dict. That is just alias the latest dates with their associated versions?

```
UNITE_DOIS["10.0"] = UNITE_DOIS["2025-02-19"]
UNITE_DOIS["9.0"] = UNITE_DOIS["2023-07-18"]
```
Or is there a better way?

Given that... we'd have mixed choices. That is, versions and dates:

```
Choices('2025-02-19', '2024-04-04', '2023-07-18',
    '2022-10-16', '2021-05-10', '2020-02-20', '10.0', '9.0' )
```

We could just add this under ❗**Breaking Changes**❗  for the [2025.4 announcement](https://forum.qiime2.org/t/preview-qiime-2-2025-4-development-changelog/31781). 🤔

nbokulich:
Yeah that is pretty messy to have a mix... sure, let's go for the breaking changes announcement. Feel free to merge whenever you are ready!

PR #220, opened 2025-04-02
add orient_reads action

### LINKED ISSUE #159: orient-seqs: accept FASTQ data as input
`vsearch --orient` can accept FASTQ as input (and also output via the `--fastqout` option). Ideally, `orient-seqs` (which is just thinly wrapping `vsearch --orient`) could do the same.

This would require modifying the inputs/outputs here: https://github.com/bokulich-lab/RESCRIPt/blob/master/rescript/orient.py#L55

HOWEVER, the main issue I see is that the current inputs and outputs are `DNAFASTAFormat` objects. A FASTQ-formatted input (e.g., coming from some of the `SampleData[.*Sequence.*]` types) could not have `DNAFASTAFormat` as a view type. I suppose we need something like a `Union[SingleLanePerSamplePairedEndFastqDirFmt | DNAFASTAFormat | ... ]` as input and output, and a `TypeMap` in the plugin registration to accept and output the corresponding types.

nbokulich:
fixes #159 

Introduces a new action, `orient-reads` to orient (FASTQ) read data, both single and paired end. Like `orient-seqs`, it can also do a simple reverse-complement if reference sequences are not provided as a template for orienting. Both modes use `vsearch` under the hood.

@mikerobeson would you mind reviewing and testing with some real data? I have made and tested single and paired-end test data, but this could use a little more rigorous user testing if you don't mind.


mikerobeson:
HI @nbokulich, this is amazing! 🤯 

I was thinking that making this a separate action would be much easier. 🙌 

Anyway, I tested this out on a decently sized data set ~ 22 samples with paired-end reads. I used with and without a reference and the output were as expected. I also tested with single-end reads too.

I reverse complimented the SILVA 138.2 database and used that as a reference to reverse orient my reads in the opposite direction. It worked quite well... on average only ~ 50 reads per sample could not be oriented.

One thing I noticed about the vsearch orient command... it displays the following warning when using `--verbose`:
>WARNING: The orient command does not support multithreading.

Which make me think that we should remove the `--p-threads` option? Using the `--dbmask` did not seem to make use of extra threads either. Unless the thinking is to later allow multiple samples to be processed concurrently? Right now this action simply iterates on one sample at a time. Which I think is fine, but wonder if this is something we should make note of for a future enhancement for both `orient-seqs` and `orient-reads`?

Otherwise, I tink it is good to merge as is...

nbokulich:
oof good call on the threads, I did not notice that. What about `--fastx_revcomp` (which is used for reverse complementing if a reference is not given for orienting seqs), does this support multi-threading? If not, I agree the param should be dropped for now.

Either way, yes I agree it would be ideal to support sample parallelization in the future, this could be done with parsl as right now this is looping through the fastqs one sample at a time.

Of note: we have the `threads` parameter exposed for `orient-seqs`, but I think that should be fixed in a separate PR for the sake of clarity.

I will open separate issues for these.

nbokulich:
Thanks so much @mikerobeson ! I dropped `threads` in the latest commit, and opened up two issues to address the other points.

mikerobeson:
LGTM! I just retested. Everything works. :-)

PR #223, opened 2025-04-03
orient-seqs: drop defunct threads param

### LINKED ISSUE #222: BUG: `orient-seqs` does not support multi-threading
## Problem
Evidently `orient-seqs` does not support multi-threading. See:
https://github.com/bokulich-lab/RESCRIPt/pull/220#issuecomment-2773043516

## Solution
Drop the `threads` parameter from this action. Could be added back with issue #221.

nbokulich:
fixes #222


mikerobeson:
Just tested to confirm. LGTM!

PR #225, opened 2025-04-03
fix orient-reads: PE and joined-pairs mode

nbokulich:
Some fixed are introduced for `orient-reads`:
* drop support for single-end reads (as the case of mixed-orientation SE reads is rather hopeless unless if the read length equals the target length)
* paired-end reads should only re-orient (swap R1/R2) but NOT reverse complement
* reverse-complement is only done for joined pairs

Thanks @mikerobeson for pointing out that the reverse complement should NOT be done for paired-end reads.


rescript/orient.py
```diff
@@ -100,11 +102,12 @@ def _vsearch_revcomp_fastq(seqs_fp, out_fp):
...
-    # adapted from q2-demux
     fh = gzip.open(filepath, 'rt')
-    for header, seq, qh, qual in zip_longest(*[fh] * 4):
-        yield (header.strip(), seq.strip(), qh.strip(), qual.strip())
+    # vsearch strips off the lane info, so we will do the same for consistency
+    for head, seq, qh, q in zip_longest(*[fh] * 4):
+        yield (head.strip().split(' ')[0], seq.strip(), qh.strip(), q.strip())
 
 
 def orient_reads(
```
mikerobeson, line 113: Set `reference_sequences: DNAFASTAFormat = ` line. At the moment `--i-reference-sequences` appears as an optional setting in the plugin.

mikerobeson (commented):
Just fix the function signature in `orient_reads`, so that `--i-reference-sequences` is required. Otherwise, everything appears to work as intended. Looks like it's ready to merge!

nbokulich:
thanks @mikerobeson ! Ref-seqs are now optional.

mikerobeson:
I re-ran on some samples, as well as the test code. It all works and LGTM! Thanks @nbokulich!

PR #229, opened 2025-04-16
ENHL adds ability to toggle GTDB download url.

### LINKED ISSUE #228: ENH: Allow user to toggle between primary and mirror GTDB urls
At times the download rate from the primary GTDB url is very slow. I think it'd bee a good idea to expose an option to toggle between the primary and mirror urls as listed [here](https://gtdb.ecogenomic.org/downloads).

That is, add a parameter called `--p-url-type` with the choices : `(['Primary', 'Mirror'])`

mikerobeson:
Fixes #228.

If downloads from the primary GTDB host site are slow, *i.e.* `--p-url-type 'Primary'`, user can switch to using the GTDB mirror site by using `--p-url-type 'Mirror'`

### DIFF
--- a/rescript/get_gtdb.py
+++ b/rescript/get_gtdb.py
@@ -36,11 +36,13 @@ def get_gtdb_data(
     version: str = '226.0',
     domain: str = 'Both',
     db_type: str = 'SpeciesReps',
+    url_type: str = 'Primary'
         ) -> (TSVTaxonomyFormat, DNAFASTAFormat):
 
     queries = _assemble_queries(version=version,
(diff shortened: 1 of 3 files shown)


nbokulich (approved):
Thanks @mikerobeson ! I tested and all is working — fwiw, the download from Primary took 40 m and from the Mirror only 1 m!

My one comment but just a thought not to fix in this PR: as we have an increasing number of releases and now the mirror, I wonder if we should add a unit test that will iterate through all possible URLs just to check that valid URLs exist for all releases, in case at some point older releases start getting deprecated.

PR #230, opened 2025-06-04
ENH: adds `get-midori2-data` action to fetch data from MIDORI2

mikerobeson:
This PR will enable users to fetch pre-compiled and curated sequence taxonomy databases from [MIDORI 2](https://www.reference-midori.info/). See their [paper](https://doi.org/10.1002/edn3.303).

The `get-midori2-data` action will only allow the ability to fetch DNA sequences from MIDORI2. I think adding a separate action to fetch amino-acid data will might come later, perhaps as `get-midori2-aa-data`?

The user will be able to specify the following parameters to fetch the pre-curated mtDNA gene sets: 
 * `mito-gene` parameter:
    * required 
    * Can choose from genes in this list: `['A6', 'A8', 'CO1', 'CO2', 'CO3', 'Cytb', 'ND1', 'ND2', 'ND3', 'ND4L', 'ND4', 'ND5', 'ND6', 'lrRNA', 'srRNA']`
        * Update (2025/06/11): User can now download a list of specified genes. Added `all` to the gene list, if the user would like to download all of the marker genes.
 * choose `version`
     * from this list: `[GenBank265_2025-03-08', 'GenBank264_2024-12-14', 'GenBank263_2024-10-13', 'GenBank262_2024-08-16', 'GenBank261_2024-06-15', 'GenBank260_2024-04-15']`
 * fetch different reference sequence types (`ref-seq-type`):
     * `uniq` : contains all unique haplotypes associated with each species
     * `longest` : contains the longest sequence for each species
 * toggle to download `unspecified-species`
     * Within the given options above, the user can also toggle if they want to download data in which there are unspecified species. That is, download any reference sequences, even if they lack binomial species-level description.

To do:
- [x] add test code.


rescript/plugin_setup.py
```diff
@@ -1106,6 +1111,48 @@
...
+    function=get_midori2_data,
+    inputs={},
+    parameters={
+        'version': Str % Choices(
+                    ['GenBank265_2025-03-08', 'GenBank264_2024-12-14',
+                     'GenBank263_2024-10-13', 'GenBank262_2024-08-16',
+                     'GenBank261_2024-06-15', 'GenBank260_2024-04-15']),
+        'mito_gene': Str % Choices(['A6', 'A8', 'CO1', 'CO2', 'CO3', 'Cytb',
+                                    'ND1', 'ND2', 'ND3', 'ND4L', 'ND4',
+                                    'ND5', 'ND6', 'lrRNA', 'srRNA']),
```
VinzentRisch, line 1124: I dont know how these sequences would be used, but is there a use case where a user would need multiple of these sequences or maybe even all?
If there is an, "all" option could be added and the input converted to a list to make it possible to download multiple  sequences at once.

rescript/get_midori2.py
```diff
@@ -0,0 +1,105 @@
...
+                        'vernum': vernum,
+                        'mito_gene': mito_gene}
+
+    fasta_url = base_url.format(**base_format_dict, **{'file_type': 'fasta'})
+    tax_url = base_url.format(**base_format_dict, **{'file_type': 'taxon'})
+
+    return fasta_url, tax_url
+
+
+def _retrieve_data_from_midori2(fasta_url, tax_url):
```
VinzentRisch, line 56: Please add tests for this function.

VinzentRisch (changes requested):
Hey @mikerobeson 
I ran it and it works without problems. See some suggestions below.

rescript/plugin_setup.py
mikerobeson: I think that this would be more appropriate as a separate QIIME 2 pipeline. Otherwise, we'd always end up with many blank QZA files for all the other marker genes that the user would not need, especially if they only wanted 1 gene.

rescript/get_midori2.py
mikerobeson: I'll add a unit test, but I am not sure if the way I am approaching it will be the correct way to do it.

rescript/plugin_setup.py
VinzentRisch: No I mean the output would be List[FeatureData[Sequence]] and List[FeatureData[Taxonomy]] and then you would just get two artifacts per chosen gene.

VinzentRisch: Any collating action is doing this. See here: https://github.com/qiime2/q2-types/blob/553bf8ea60904e692b78d87cbc795beeb0acdccf/q2_types/per_sample_sequences/_methods.py#L63-L100

VinzentRisch: Hey @mikerobeson 
I am so sorry, I gave you a wrong and confusing example! It's the partition actions that have the Collection that i wanted to show you. So here you can see you can register an output as a collection: https://github.com/qiime2/q2-types/blob/553bf8ea60904e692b78d87cbc795beeb0acdccf/q2_types/plugin_setup.py#L53
That means that the output are any number of artifacts of the same type. 
And the parameter 'mito_gene' would then be registered as a List in the plugin setup. I think like this:
```suggestion
        'mito_gene': List[Str % Choices(['A6', 'A8', 'CO1', 'CO2', 'CO3', 'Cytb',
                                    'ND1', 'ND2', 'ND3', 'ND4L', 'ND4',
                                    'ND5', 'ND6', 'lrRNA', 'srRNA', 'all'])],
```
And then in the action you loop through the list of parameters and for every parameter that is given by the user your logic runs and creates an artifact that is added to a dict and then returned: 
https://github.com/qiime2/q2-types/blob/553bf8ea60904e692b78d87cbc795beeb0acdccf/q2_types/per_sample_sequences/_methods.py#L58-L60

If 'all' is given then it loops through all the possible gene names. 

So I am giving the partition and collation actions as examples because they use these List and Collection classes in the plugin setup but you dont need any of the partition or collation logic. So no you dont have to run it in two parts. 

I'm sorry for the confusion and for wasting your time. 😬 
It should be straight forward now but will take some time so if you dont have time you can also add it as a new issue.

rescript/tests/test_get_midori2.py
```diff
@@ -0,0 +1,111 @@
...
+                 'Databases/GenBank263_2024-10-13/QIIME_sp/longest/MIDORI2_'
+                 'LONGEST_NUC_SP_GB263_CO1_QIIME.taxon.gz')
+        self.assertEqual(fasta_url, exp_fasta_url)
+        self.assertEqual(tax_url, exp_tax_url)
+        # test that these URLs work
+        for u in [fasta_url, tax_url]:
+            try:
+                urlopen(u)
+            except HTTPError:
+                raise ValueError('Failed to open URL: ' + u)
```
VinzentRisch, line 61:
```suggestion
```
This unittest should only test your logic, so only the assembly of the string.

rescript/get_midori2.py
```diff
@@ -0,0 +1,103 @@
...
+                                             mode="r").view(pd.DataFrame)
+    return seqs, tax
+
+
+def get_midori2_data(
+    mito_gene: str,
+    version: str = 'GenBank265_2025-03-08',
+    ref_seq_type: str = 'uniq',
+    unspecified_species: bool = False,
+        ) -> (DNAIterator, pd.DataFrame):
```
mikerobeson, line 91: @VinzentRisch, assuming that I update given your other comments... I should set `mito_gene: list`. But what do I set the expected output to?  I'd imagine it can't still be `-> (DNAIterator, pd.DataFrame)`.

rescript/tests/test_get_midori2.py
mikerobeson: Yeah, we decided to do this originally for some repos, as the servers would often go down while were where working on code. I often like to leave it in, but have removed as per your suggestion.

rescript/tests/test_get_midori2.py
```diff
@@ -0,0 +1,111 @@
...
+                 'Databases/GenBank260_2024-04-15/QIIME/uniq/MIDORI2_'
+                 'LONGEST_NUC_GB260_ND4L_QIIME.fasta.gz')
+        incorrect_tax_url = (
+                 'https://www.reference-midori.info/download/'
+                 'Databases/GenBank260_2024-04-15/QIIME/uniq/MIDORI2_'
+                 'LONGEST_NUC_GB260_ND4L_QIIME.taxon.gz')
+
+        with self.assertRaises(ValueError):
+            _retrieve_data_from_midori2(incorrect_fasta_url,
+                                        incorrect_tax_url)
```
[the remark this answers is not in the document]
mikerobeson: I tried doing this, but failed miserably. I like this as it explicitly tests the exception error I have in the code. But I assume the side effect part will help text that directly? I could not see how the mock code would actually test the exception in my code. So, I let this be for now.

You're very good at finding the things I am not good at! 😅

rescript/get_midori2.py
mikerobeson: Got it working! 🎆

rescript/get_midori2.py
```diff
@@ -0,0 +1,117 @@
...
+            in_path = os.path.join(tmpdirname, compressed_fn)
+            out_path = os.path.join(tmpdirname, uncompressed_fn)
+
+            try:
+                print('Retrieving data from {0}'.format(url))
+                urlretrieve(url, in_path)
+            except Exception as e:
+                raise ValueError(
+                    f'''Unable to retrieve the following file
+                    from MIDORI2:\n{url}\n: {e}''') from e
```
VinzentRisch, line 75:
```suggestion
                    f'Unable to retrieve the following file from MIDORI2:\n{url}\n{e}'
                ) from e
```
The trippel quotes introduce an unwanted new line.

rescript/tests/test_get_midori2.py
```diff
@@ -0,0 +1,103 @@
...
```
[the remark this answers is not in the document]
mikerobeson: Thank you for this! Much more simple than I had realized. I fixed it up a bit. It works! 🙏

Issue #233, opened 2025-07-11
ENH: add ability to fetch PR2 version 5.1.0


mikerobeson:
Recently, PR2 released version [5.1.0](https://github.com/pr2database/pr2database/releases/tag/v5.1.0.0).

To do:

Add `5.1.0` as an option for `--p-version`

PR #235, opened 2025-07-11
ENH: updating and refactoring `get-pr2-data`

### LINKED ISSUE #234: BUG: get-pr2-data will fail when pulling `4.14.0`
Need to either remove `v4.14.0` from the allowed list of reference db versions to fetch, or refactor the code. The rank `Supergroup` is not available.

This version was added by mistake, as `get-pr2-data` was only intended to fetch versions that support 9-ranks, *i.e.* versions `5.0.0` and later. Using version `4.14.0` will fail as the code expects 9 available ranks, when there are only 8 available ranks.

mikerobeson:
TL;DR: This PR fixes #234 and #233.

- Version `4.14.0` removed.
  - Adding this was a mistake, as this version only has 8 ranks, while the code expects 9 ranks.
    - That is, the rank `Supergroup` is not available.
  - See #234. 
- Version `5.1.0` added
  - This means that two versions are now available: `5.0.0` & `5.1.0`. 
  - Due to minor changes in the URL format, I decided to make a dictionary of links for each version.
    - See #233.  
  - This will make it easier if any other non-standard URL links are generated. That is, no need to write complicated parsers to derive the URL for a given version.
 
- [ ] Complete Test Code


mikerobeson:
Thanks for reviewing my PRs @VinzentRisch! For this one, I decided to refactor the code quite a bit for easier testing. I think I like were I landed... I'll work on adding more test code in the coming week.

PR #252, opened 2026-02-19
FIX: correctly set the DB param before the EFetch request

misialq:
When using `query` to fetch protein sequences we would make two requests: esearch + efetch. The DB for the first was set correctly, while for the second one it was hardcoded to "nuccore" (which is incorrect for fetching protein sequences). The code was updated to dynamically set the DB for the second request, depending on the DB used for the first request.

Note: I'm not sure why this worked before but perhaps NCBI just started enforcing passing the correct DB in the linked EFetch request...

### DIFF
--- a/rescript/ncbi.py
+++ b/rescript/ncbi.py
@@ -243,7 +243,7 @@ def _esearch(params, logging_level, entrez_delay=0.334):
         if 'WebEnv' not in webenv:
             raise ValueError('No sequences for given query')
         params = dict(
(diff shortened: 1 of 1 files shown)


mikerobeson:
Thank you for doing this @misialq! I was able to successfully run the command as outlined from [this user's post](https://forum.qiime2.org/t/rescript-get-ncbi-genomes-and-get-ncbi-data-protein-functions-not-supported/33995). All tests also passed. I'll go ahead and merge. 🙏

PR #254, opened 2026-05-04
added GTDB v232 as available download option

### LINKED ISSUE #253: ENH: add ability to download GTDB 232
GTDB [released](https://forum.gtdb.ecogenomic.org/t/announcing-gtdb-r11-rs232/826) version 232. This would be great to add to the list of downloadable versions. Once that's done, I can train classifiers for it (or if it will take a little while, I can pull the files directly - but it's nice to have the RESCRIPt download in provenance). 

Side note: From their release announcement:

> Changed methodology for identifying and extracting 16S rRNA genes to mirror Prokka (barnap), including setting minimum length to 396 bp.

I've been poking around with the GTDB 226 and `feature-classifier extract-reads` and just noticed today that there are about 10,000 sequences in the GTDB 226 16S reference that are less than 396 bases long.

mikerobeson:
Fixes #253.


gregcaporaso:
Thanks @mikerobeson!

Issue #265, opened 2026-09-10
ENH: Add ability to download SILVA sequence alignment files via `get-silva-data`.


mikerobeson:
At the moment we only offer the downloading of unaligned SILVA sequence files via `get-silva-data`. Although there is a "manual" work around as outlined [here](https://forum.qiime2.org/t/train-a-taxonomy-classifier-by-extracting-an-amplicon-region-from-a-curated-sequence-alignment-using-silva-as-an-example/34289), it'd be nice to simply include the ability to download the sequence alignment files as an option.

I am thinking that simply re-configuring `get-silva-data` to make use of `Collection[FeatureData[Sequence]]` so that unaligned and/or aligned sequence data can be pulled. Basically, follow the output strategy similar to `get-midori2-data` and `get-eukaryome-data`.  Though I think that `Collection` won't work this way. If not, I'll just slightly update the way it currently works.

I am thinking that the collection could be set as optional, which would obviate the need for the `--p-download-sequences` option?

mikerobeson:
closed via #266

PR #266, opened 2026-09-11
ENH: Allow the ability to download SILVA alignment.

mikerobeson:
# Description

Addresses #265. 
Added the flag `--p-seq-format`. This will allow the user to download :
- `Aligned` sequence data
- `Unaligned` sequence data
- `Both` the aligned and unaligned sequence data
- `None` of the sequence data, just the taxonomy

# AI Disclosure

- [x] NO AI USED.
- [ ] AI USED.

# AI Usage Details

### DIFF
--- a/rescript/get_data.py
+++ b/rescript/get_data.py
@@ -42,19 +42,12 @@ def get_silva_data(ctx,
         ranks=ranks,
         rank_propagation=rank_propagation)
     # if skipping sequences, need to output an empty sequence file.
-    if seq_format == 'None':
-        results['sequences'] = qiime2.Artifact.import_data(
-            'FeatureData[RNASequence]', RNAFASTAFormat())
+    if seq_format == 'Unaligned' or seq_format == 'None':
         results['aligned sequences'] = qiime2.Artifact.import_data(
             'FeatureData[AlignedRNASequence]', AlignedRNAFASTAFormat())
-    if seq_format == 'Unaligned':
-        results['aligned sequences'] = qiime2.Artifact.import_data(
-            'FeatureData[AlignedRNASequence]', AlignedRNAFASTAFormat())
-    if seq_format == 'Aligned':
+    if seq_format == 'Aligned' or seq_format == 'None':
         results['sequences'] = qiime2.Artifact.import_data(
             'FeatureData[RNASequence]', RNAFASTAFormat())
-    # else:
-    #     seq_res = {k: results[k] for k in ['sequences', 'sequences_aligned']}
     return results['sequences'], results['aligned sequences'], taxonomy
 
 
--- a/rescript/get_data.py
+++ b/rescript/get_data.py
@@ -58,7 +58,7 @@ def get_silva_data(ctx,
     return results['sequences'], results['aligned sequences'], taxonomy
 
 
-def _assemble_silva_data_urls(version, target, seq_format): # download_sequences=True
+def _assemble_silva_data_urls(version, target, seq_format):
     '''Generate SILVA urls, given database version and reference target.'''
     # assemble target urls
     ref_map = {'SSURef_NR99': 'ssu_ref_nr',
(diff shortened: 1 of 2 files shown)


rescript/get_data.py
```diff
@@ -25,11 +25,12 @@ def get_silva_data(ctx,
                    include_organism_name_labels=False,
                    rank_propagation=True,
                    ranks=None,
-                   download_sequences=True):
+                   seq_format='Unaligned',
+                   # download_sequences=True,
```
VinzentRisch: The download_sequences parameter should not be removed but rather deprecated. if download_sequences is set to True then internally seq_format is set to The option that would preserve the same outcome as before the PR and a warning is being printed informing that this parameter is deprecated and will be removed in the next release with a class like this:

`class RachisFutureWarning(RachisWarning, FutureWarning):`
Also a check is needed to prevent using download_sequences and seq_format at the same time.

rescript/plugin_setup.py
```diff
@@ -719,8 +719,10 @@
         'include_organism_name_labels': Bool,
         'rank_propagation': Bool,
         'ranks': List[Str % Choices(ALLOWED_RANKS)],
-        'download_sequences': Bool},
+        'seq_format': Str % Choices(['Aligned', 'Unaligned', 'Both', 'None'])
```
VinzentRisch: The parameter choices should be lowercase.

VinzentRisch (changes requested):
Hey @mikerobeson 
It would be better if the download_sequences parameter is not directly removed but rather deprecated. See below for changes.

rescript/get_data.py
mikerobeson: Thanks @VinzentRisch! I was thinking this would fit under "breaking changes" in the release notes. But I see your point. 

Though this will be confusing to users as they will see `download_sequences=False` as default, while `seq_format='Unaligned'`, with the help text contradicting each other, despite the deprecation warning.

If we want to do the deprecation route, I think we should always bring up the deprecation notice, and leave `download_sequences=True`. Then as long as `seq_format` is not `none`, we're okay. Otherwise, with the conflicting selections between the two parameters, we can break out of the action by raising an error and print the deprecation notice. 

I think that it'd be easiest to use a TypeMap for this, as it should already handle unwanted combinations. It also handle breaking out of the action. Both of these would save extra code writing.

I must be blind, I am unable to find an example in the code base for using `RachisFutureWarning`? Can you link an example?

VinzentRisch: Hey @mikerobeson 
Sorry I got confused in my earlier comment. Yes your approach sounds good! 
The RachisFutureWarning does not exist, it is just a way to raise a warning that is a FutureWarning and a RachisWarning at the same time. If you only use FutureWarning, I think the warning will only be displayed when a user sets `--verbose`.

mikerobeson: Hi @VinzentRisch, I've added `TypeMap` as described above. I also updated the help text to highlight depreciation of `download_sequences`.  However, I am unable to get warnings to work. I suspect I am not implementing this correctly. It'd be cool if we had something similar to the deprecation annotation that we can supply in the `plugin_setup.py` but not for notifying about the entire plugin, but aspects of the plugin, *e.g.* actions that will change or be deprecated. Any thoughts on this?

VinzentRisch: Hey @mikerobeson 

I had to remove the type map again. I think this is better because existing commands continue to behave the same during the deprecation period. What do you think?
I now just used RachisWarning. The combination does not work with FutureWarning.

The test test_retrieve_data_from_silva is failing could you inspect that?

mikerobeson: Yeah, you're correct. This is much cleaner than I was thinking. I never knew about `CaptureHolder` and `IContext`. I learned something new. Thank you! 🙏 

Yep I'll take a look at `test_retrieve_data_from_silva`.

mikerobeson: Hey @VinzentRisch,  the url  for the file being tested in `test_retrieve_data_from_silva` had moved / changed. So I updated the url. All the tests pass now. If you're okay with the PR, I can merge.

VinzentRisch: Hey @mikerobeson 
Yes CaptureHolder is new and quite handy with random seeds for instance. 
I didn't test all the combinations for the silva downloads with the new parameter handling. Did you test the combinations? If yes you can merge if not could you do that?

mikerobeson: Yep, I already tested your parameter combinations, they fail and work as expected. :

VinzentRisch: Great then feel free to merge :)

